PR #6, opened 2024-10-05
Update train_gpt2.py

somefunAgba:
make gradient smoothing unbiased.

@KellerJordan could you try this and see how if it improves the performance.


KellerJordan:
It won't matter, since newton_schulz is invariant to the scale of its input

somefunAgba:
I suggest you try it though and see if your intuition is right or not.

PR #16, opened 2024-11-07
15% Faster Implementation of Muon

leloykun:
This PR contains an implementation of Muon w/ the minimal operations. Benchmarks are provided.

Caveat: this only decreases total training time for the 125M model by < 1s. But the results would be more pronounced on larger models so it's worth adding this in anyway. See [training logs](https://github.com/leloykun/modded-nanogpt/blob/fc--compiled-muon/logs/015f6462-7563-4bce-960e-1b47f06ab425.txt)

---

Original implementation:
```python
A = X @ X.T
B = A @ X
X = a * X + b * B + c * A @ B
```

New implementation:
```python
A = X @ X.T
S = A @ (b * I + c * A)
torch.diag(S).add_(a)
X = S @ X
```

image

---

Benchmark results:

Original implementation: `0.2644890202204944 ms`
This implementation:       `0.2234299657944715 ms`

So, around 15% faster on average.

---

c: @youjiacheng did initial algebraic rewrites, I further reduced ops, then @youjiacheng proposed the `torch.diag` trick


zhangfaen:
looks cool

KellerJordan:
This definitely looks interesting, although I'm not sure yet that it's faster for the actual speedrun, which is what this particular repo is focused on

KellerJordan:
Looks interesting, and was suggested by jxbz before as well, but I'm just not sure yet that this is actually faster for the speedrun, which is what this repo's focused on

KellerJordan:
Closing because I've now made this change, which is similar: https://github.com/KellerJordan/modded-nanogpt/commit/ab0eb61af1280ae2260dd6fec35d5a49deab112d

The earliest person to suggest something like this was @jxbz like a month ago

In my tests, adding a to the diagonal was somewhat slower than what I ended up going with.

Btw I think the right way to add to diagonal would be S.diagonal().add_(a), using torch.diag instead seems to make a separate/cloned tensor

KellerJordan:
granted contributor status in the Muon writeup for this! https://kellerjordan.github.io/posts/muon/

Issue #24, opened 2024-11-11
Cant run any reproductions when using 8xH100 SXM5


bluecoconut:
I've now tried this on different vendors and keep getting errors.

Most recent version is:
I try and run on lambdalabs 8xH100 SXM5, and it fails: https://github.com/KellerJordan/modded-nanogpt/issues/15

so I tried the fix there: `pip install --upgrade nvidia-nccl-cu1`

but im now getting other errors. I've tried with both the most recent record (`111024_UNetDoubleLR` as well as the `train_gpt2.py` thats checked into master).

```
ubuntu@192-222-52-59:~$ nvidia-smi
Mon Nov 11 20:00:18 2024       
+-----------------------------------------------------------------------------------------+
| NVIDIA-SMI 550.90.12              Driver Version: 550.90.12      CUDA Version: 12.4     |
|-----------------------------------------+------------------------+----------------------+
| GPU  Name                 Persistence-M | Bus-Id          Disp.A | Volatile Uncorr. ECC |
| Fan  Temp   Perf          Pwr:Usage/Cap |           Memory-Usage | GPU-Util  Compute M. |
|                                         |                        |               MIG M. |
|=========================================+========================+======================|
|   0  NVIDIA H100 80GB HBM3          On  |   00000000:61:00.0 Off |                    0 |
| N/A   25C    P0             69W /  700W |       1MiB /  81559MiB |      0%      Default |
|                                         |                        |             Disabled |
+-----------------------------------------+------------------------+----------------------+
|   1  NVIDIA H100 80GB HBM3          On  |   00000000:62:00.0 Off |                    0 |
| N/A   27C    P0             69W /  700W |       1MiB /  81559MiB |      0%      Default |
|                                         |                        |             Disabled |
+-----------------------------------------+------------------------+----------------------+
|   2  NVIDIA H100 80GB HBM3          On  |   00000000:63:00.0 Off |                    0 |
| N/A   27C    P0             70W /  700W |       1MiB /  81559MiB |      0%      Default |
|                                         |                        |             Disabled |
+-----------------------------------------+------------------------+----------------------+
|   3  NVIDIA H100 80GB HBM3          On  |   00000000:64:00.0 Off |                    0 |
| N/A   26C    P0             72W /  700W |       1MiB /  81559MiB |      0%      Default |
|                                         |                        |             Disabled |
+-----------------------------------------+------------------------+----------------------+
[...]

Issue #15, opened 2024-11-07
Error running on 8xH100, works on 4xH100?


swookey-thinky:
Am attempting to reproduce your results, as was wondering if you have hit the following issue before. I am using a Lambda Labs 8xH100 SXM5 instance, and run the following commands from a fresh instance of the repository:

```
python -m venv nanogpt_env
source nanogpt_env/bin/activate
pip install --upgrade pip
pip install -r requirements.txt
python data/cached_fineweb10B.py 18
./run.sh
```

On a 4xH100 SXM5 instance from Lambda, everything runs fine and I can successfully reproduce everything. Similarly on an 8xA100 instance, it works as well. However, solely on the 8xH100 instances, training fails in the child processes with:

```
[I1107 01:20:27.770733060 TCPStoreLibUvBackend.cpp:119] [c10d - debug] Read callback failed. code:-4095 name:EOF desc:end of file
```

Have you seen this issue before? I suspect a NCCL issue but I have never run across this issue before. Here is the full set of logs with detailed debugging:

```
[I1107 01:20:06.503069695 debug.cpp:49] [c10d] The debug level is set to DETAIL.
I1107 01:20:06.697000 7974 torch/distributed/run.py:897] 
I1107 01:20:06.697000 7974 torch/distributed/run.py:897] **************************************
I1107 01:20:06.697000 7974 torch/distributed/run.py:897] Rendezvous info:
I1107 01:20:06.697000 7974 torch/distributed/run.py:897] --rdzv-backend=c10d --rdzv-endpoint=localhost:0 --rdzv-id=c0d436eb-46c3-4d17-8205-20cd784b1819
I1107 01:20:06.697000 7974 torch/distributed/run.py:897] **************************************
I1107 01:20:06.697000 7974 torch/distributed/run.py:897] 
W1107 01:20:06.697000 7974 torch/distributed/run.py:793] 
W1107 01:20:06.697000 7974 torch/distributed/run.py:793] *****************************************
W1107 01:20:06.697000 7974 torch/distributed/run.py:793] Setting OMP_NUM_THREADS environment variable for each process to be 1 in default, to avoid your system being overloaded, please further tune the variable for optimal performance in your application as needed. 
W1107 01:20:06.697000 7974 torch/distributed/run.py:793] *****************************************
I1107 01:20:06.698000 7974 torch/distributed/launcher/api.py:194] Starting elastic_operator with launch configs:
[...]

PR #25, opened 2024-11-12
Dockerfile and update train_gpt2.py to most recent record

bluecoconut:
I specifically tried to reproduce the fastest run on 2 different 8xH100 SXM5 vendors and ran into lots of issues; installing various versions of cuda, pytorch, etc. 

In the end just standardizing around a docker run feels like the best solution since iterating on cuda installs takes a long time and depending on the base-OS-stack (eg. lambda stack, etc.) can produce a broken system. Additionally, this ensures that the speedrun is more "pinned" to specific versions of cuda, cudnn, python, ubuntu, etc. 

This PR introduces a dockerfile that sets up the most up-to-date python version (3.12.7), most up to date LTS version of ubuntu (24.04), and current up-to-date cuda (12.6.2)

I also included some updates to the readme, mapping the same "3" commands (install, download data, run)

~~Additionally, since I wanted to be sure I was running the actual current record, I copied the code from the `/records` folder into `train_gpt2.py`~~
Edit: Decided to remove this, it seems like most of the change from train_gpt.py and the output in records/... is "print0" style file + console output vs. split file + console output, and that might be under active work -- so I don't want to force any unnecessary or unintend change.

One small extra fix, for the docstring on `zeropower_via_newtonschulz5` I added `r"` to suppress the `\s` warning by using raw-string literal.


bluecoconut:
this i think resolves issues such as:
https://github.com/KellerJordan/modded-nanogpt/issues/24
https://github.com/KellerJordan/modded-nanogpt/issues/15

KellerJordan:
After having some problems on a new vendor, yeah I agree this is a good idea.

PR #27, opened 2024-11-13
Replace iframe tag with image link to `Initial D - Deja vu`

dantetemplar:
image
Currently, github users see just html tag text without rendering. I guess it will be better to include the thumbnail with link.

Actually, I love this anime :3


KellerJordan:
tru

PR #28, opened 2024-11-21
Optimize data loading (saves ~1ms per step on my runs)

leloykun:
A minor change, but I'm raising a PR so more people can double check it

Changelogs:
- Directly casts into `np.int64` instead of casting to `np.int32` first then to `torch.long`
- Sets `pin_memory=True` so that data would be loaded directly into the GPU without going thru the CPU first
- Loads data to CUDA asynchronously 

I could also add prefetching and such... but that would be overkill


leloykun:
nvm, it might even be making things worse

closing for now...

PR #31, opened 2024-11-25
Correcting gradient accumulation in the training section

timlautk:
Putting the forward pass within the no_sync() context along with the backward pass; otherwise gradients will still be synchronized. For details, check this from the PyTorch repo: https://github.com/pytorch/pytorch/blob/main/torch/nn/parallel/distributed.py#L1427


KellerJordan:
this might be slightly slower due to moving next_batch after the backward. not sure will have to check

timlautk:
You're right. I have check that. Just putting that back before the backward.

Issue #29, opened 2024-11-22
A speedrun on consumer grade cards?


fzyzcjy:
Hi thanks for the great repo! I would appreciate it if there can be a speed run on consumer cards e.g. RTX4090. Since it is 125M params, the RTX4090's 24GB memory should fit in the classical way, and thus it is trainable.

KellerJordan:
a suggestion, to reduce memory you could run with a lower sequence length

fzyzcjy:
I think so, thanks :) Just wondering whether there will be a speedrun like the current great one but focus on RTX4090's time, because many more people have consumer grade cards than H100s.

naoro:
I think a Google colab speed runs would be also awesome -
That would greatly commodotize research and experimentation

fzyzcjy:
That looks interesting! (But I guess it may be too hard to have it run in acceptable times...)

alexjc:
Realistically single-card speed-run would need a smaller model too, otherwise it's too slow to experiment.

Thinking:
```
n_layer = 8
n_embd = 512
```

The sequence length during training has been a variable factor in the last speed run; for evaluation it's fine if it's the whole document clamped at maximum 1024 window size?

fzyzcjy:
It seems H100 has 2000 TFLOPS for bf16 tensor core, while 4090 is about 330 TFLOPS. Thus 8xH100 5minute = 1x4090 4hour, which looks not bad!

The major problem looks like the memory is only 24GB... So we may not be able to do some optimizations.

alexjc:
4h feels quite high for a speed run. Too hard to test ideas, no?

Working on some memory optimizations now, should help a lot...

fzyzcjy:
Surely faster would be great! But if impossible, then 4h is better than nothing :(

naoro:
> It seems H100 has 2000 TFLOPS for bf16 tensor core, while 4090 is about 330 TFLOPS. Thus 8xH100 5minute = 1x4090 4hour, which looks not bad!
> 
> The major problem looks like the memory is only 24GB... So we may not be able to do some optimizations.

A100 has got 40GB, cost is about 10usd for ~12 hours with about the same tflops as 4090.
So about 3.3usd per run. Not bad, I'd think.

fzyzcjy:
If we optimize for cost, then 4090 is much cheaper than A100 per hour while having same tflops. Thus as long as we manage to fit in 24GB, then maybe we can further scale down the cost.

[...]

PR #38, opened 2024-11-29
Run on RTX 4090 

lapp0:
Changes to make the code run on RTX 4090 / 3090.

Fixes https://github.com/KellerJordan/modded-nanogpt/issues/29

[Runs in 2 hours 3 minutes](https://gist.github.com/lapp0/ff6f10c3cd6d0aefb28a49681a44b78c), Runs range from 3.275 to 3.285. This finished at 3.2817, 

These settings are intended to replicate the training dynamics on 8xH100, not to be optimal. This is accomplished by halving the sequence length and doubling the batch size. You can train in ~90 minutes by setting batch_size of 1.

https://github.com/KellerJordan/modded-nanogpt/pull/38#issuecomment-2518255097


KellerJordan:
I will have to think about this and understand the options before merging this.

One thing I would merge immediately would be a reproducible log generated by this. I could be put in the `records` folder and linked to in the README for people who wanna train on 4090. I might prefer that, since it would avoid adding options to the code.

lapp0:
Perhaps a reasonable alternative is
- 1) Expose hyperparameters as command arguments via argparse
- 2) Document the command which enables 4090 trains

lapp0:
@vgoklani I stole the values from https://github.com/pytorch/pytorch/issues/133254#issuecomment-2408710459

banyan-god:
> Perhaps a reasonable alternative is
> 
>     * 1. Expose hyperparameters as command arguments via argparse
> 
>     * 2. Document the command which enables 4090 trains

definitely prefer this over how it is implemented now, especially if you want to use multiple 4090 and lets say we want to keep the context length same but change the batch size instead .

lapp0:
Updated code and docs. You can now specify `GPTConfig` and `Hyperparameters` as command line arguments, e.g.

```
torchrun --standalone --nproc_per_node=4 train_gpt2.py \
    --gpt.flex_kernel_consumer True --train.sequence_length 32768 --train.batch_size 32
```

[Rendered Docs](https://github.com/KellerJordan/modded-nanogpt/blob/a77f7eb470ec0c9f7d0199b8c8b194a6d5ad3ebb/README.md#consumer-hardware-and-training--model-configuration)

KellerJordan:
I linked to this from the README. I don't think I'll merge it because I don't want to add any arguments to `train_gpt2.py`.

lapp0:
Perhaps we could have `train_gpt2.py` be

```
def train(model_config: GPTConfig, args: Hyperparameters):
    ...
    
if __name__ == "__main__":
    train(GPTConfig(), Hyperparameters())
 ```

This is functionally equivalent to the current code except it puts the trainer in a function parameterized by the model and training arguments, cleaning the code and making it easier to employ hyperparameter search tools.

Then we can create `train_gpt2_cli.py` which imports `train` from `train_gpt2.py` and handles arg parsing.

**Comment by vak** [comment]
@lapp0 many thanks for your fork!

I've started your fork successfully on a single RTX 4090 with minor version upgrades and a tiny fix in log-output writing:

```
➜  modded-nanogpt git:(pr-38) ✗ git diff
diff --git a/Dockerfile b/Dockerfile
index 4ca8600..57fcced 100644
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,7 +1,7 @@
-FROM nvidia/cuda:12.6.2-cudnn-devel-ubuntu24.04
+FROM nvidia/cuda:12.6.3-cudnn-devel-ubuntu24.04
 
 ENV DEBIAN_FRONTEND=noninteractive
-ENV PYTHON_VERSION=3.12.7
+ENV PYTHON_VERSION=3.12.8
 ENV PATH=/usr/local/bin:$PATH
 
 RUN apt update && apt install -y --no-install-recommends build-essential libssl-dev zlib1g-dev \
@@ -27,7 +27,7 @@ WORKDIR /modded-nanogpt
 RUN python -m pip install --upgrade pip && \
 
-RUN pip install --pre torch --index-url https://download.pytorch.org/whl/nightly/cu124 --upgrade
+RUN pip install --pre torch --index-url https://download.pytorch.org/whl/nightly/cu126 --upgrade
 
 CMD ["bash"]
 ENTRYPOINT []
diff --git a/run.sh b/run.sh
index 7bb09f4..04e486d 100755
--- a/run.sh
+++ b/run.sh
@@ -1 +1 @@
-torchrun --standalone --nproc_per_node=8 train_gpt2.py
+torchrun --standalone --nproc_per_node=1 train_gpt2.py
diff --git a/train_gpt2.py b/train_gpt2.py
index 6f7f74e..35b1d87 100644
--- a/train_gpt2.py
+++ b/train_gpt2.py
@@ -437,7 +437,7 @@ def train(model_config: GPTConfig, args: Hyperparameters):
             with open(logfile, "a") as f:
                 if not logonly:
                     print(s)
-                f.write(s+'\n')
+                f.write(str(s)+'\n')
     # log information about the hardware/software environment this is running on
     # and print the full `nvidia-smi` to file
     print0(f"Running pytorch {torch.version.__version__} compiled for CUDA {torch.version.cuda}\nnvidia-smi:")
```

However I am running out of GPU memory. Probably because I use single GPU setup?

If you have an idea what parameters to tweak to fit in a single RTX 4090 24GB, please let me know, 
thank you!

lapp0:
@vak  you can always decrease sequence length further and increase batch size in the same ratio.

lapp0:
With `torch==2.6.0.dev20241231` you don't need to specify the kernel_options. Simply change the halve and double the sequence length and batch size respectively.

BradKML:
@vak and @lapp0 now I am curious how and where to do this (in the case of running these models on a laptop GPU), it is still those two places?

PR #42, opened 2024-12-05
Fix peak memory usage not being saved to the logfile

tysam-code:
Super simple, just lets us see the peak memory usage from past model runs at the end of the logfile (ran into this today, so just fixing it real quick!)


KellerJordan:
lgtm 👍

PR #44, opened 2024-12-07
Log numpy version as well as pytorch version

segyges:
This seems advisable because the dataloader uses numpy extensively and oddities may crop up later with numpy updates.


segyges:
Added logging of python version/build too, which seems advisable for similar reasons of thoroughness in case of future oddities.

PR #47, opened 2024-12-09
per-head Q, K, V gradient orthonormalization

scottjmaddox:
This doesn't appear to significantly improve the loss, but it does speed up training by ~1% (on 1xH100), by splitting the orthonormalization task into `n_head` sub-tasks. Not sure if that's significant enough to qualify as a record.

If you think it's worth pursuing I can rent an 8xH100 to collect enough runs for the t-test. Although if you or others have spare 8xH100 capacity, that would be great.

### DIFF
--- a/train_gpt2.py
+++ b/train_gpt2.py
@@ -76,16 +76,16 @@ class Muon(torch.optim.Optimizer):
         defaults = dict(lr=lr, momentum=momentum, nesterov=nesterov, ns_steps=ns_steps)
         params = list(params)
         assert all(isinstance(p, torch.Tensor) for p in params)
-        sorted_shapes = {tuple(sorted(p.shape)) for p in params}
+        ndims_and_sizes = {(p.ndim, p.numel()) for p in params}
         param_groups = [
             {
-                'params': [p for p in params if tuple(sorted(p.shape)) == sorted_shape],
+                'params': [p for p in params if p.ndim == ndim and p.numel() == size],
                 'update_buffer': [
-                    torch.empty(sorted_shape, device='cuda', dtype=torch.bfloat16).flatten()
+                    torch.empty(size, device='cuda', dtype=torch.bfloat16)
                     for _ in range(self.world_size)
                 ],
             }
-            for sorted_shape in sorted_shapes
+            for ndim, size in ndims_and_sizes
         ]
         super().__init__(param_groups, defaults)
 
--- a/train_gpt2.py
+++ b/train_gpt2.py
@@ -76,16 +76,16 @@ class Muon(torch.optim.Optimizer):
         defaults = dict(lr=lr, momentum=momentum, nesterov=nesterov, ns_steps=ns_steps)
         params = list(params)
         assert all(isinstance(p, torch.Tensor) for p in params)
-        shapes = {p.shape for p in params}
+        sorted_shapes = {tuple(sorted(p.shape)) for p in params}
         param_groups = [
             {
-                'params': [p for p in params if p.shape == shape],
+                'params': [p for p in params if tuple(sorted(p.shape)) == sorted_shape],
                 'update_buffer': [
-                    torch.empty(shape, device='cuda', dtype=torch.bfloat16).flatten()
+                    torch.empty(sorted_shape, device='cuda', dtype=torch.bfloat16).flatten()
                     for _ in range(self.world_size)
                 ],
             }
-            for shape in shapes
+            for sorted_shape in sorted_shapes
         ]
         super().__init__(param_groups, defaults)
 
--- a/train_gpt2.py
+++ b/train_gpt2.py
@@ -76,16 +76,16 @@ class Muon(torch.optim.Optimizer):
         defaults = dict(lr=lr, momentum=momentum, nesterov=nesterov, ns_steps=ns_steps)
         params = list(params)
         assert all(isinstance(p, torch.Tensor) for p in params)
-        ndims_and_sizes = {(p.ndim, p.numel()) for p in params}
+        sizes = {p.numel() for p in params}
         param_groups = [
             {
-                'params': [p for p in params if p.ndim == ndim and p.numel() == size],
+                'params': [p for p in params if p.numel() == size],
                 'update_buffer': [
                     torch.empty(size, device='cuda', dtype=torch.bfloat16)
                     for _ in range(self.world_size)
                 ],
             }
-            for ndim, size in ndims_and_sizes
+            for size in sizes
         ]
         super().__init__(param_groups, defaults)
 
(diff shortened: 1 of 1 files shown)


leloykun:
Interesting...

I got a lot worse results when I tried something like this (each line below is an average over 4 runs).

image

KellerJordan:
thanks for this contribution. it's a small claimed benefit, so I think it will be basically up to the motivation of whoever has enough GPUs and wants to try to get statistical significance for it. will leave the issue open

lapp0:
Worth mentioning that I saw a measurable reduction in step-time when I convert the QKV projections into a single linear layer, then chunked the linear out into the q,k,v components. The problem was that Muon treating this as a single parameter resulted greater loss.

scottjmaddox:
> I got a lot worse results when I tried something like this (each line below is an average over 4 runs).

@leloykun  Are you sure you made all of the necessary changes to the Muon optimizer, both the implementation and selection of which optimizer to assign to each parameter tensor? My first couple attempts showed significant degradation, but I later noticed additional changes that needed to be made. With all of the changes included in this PR, there seems to be no significant change to the loss (over a small number of runs).

I'll rebase the PR and collect some 8xH100 runs.

scottjmaddox:
> Worth mentioning that I saw a measurable reduction in step-time when I convert the QKV projections into a single linear layer, then chunked the linear out into the q,k,v components. The problem was that Muon treating this as a single parameter resulted greater loss.
>
> You might consider trying the same and chunking the grad into individual heads before orthogonalizing in Muon.

Yeah, using a single matmul rather than 3 seems to be a common optimization, but I'm not sure how to do that with pytorch's optimizer API and Muon's semantics. I suppose you could add a custom 3-dim optimizer used only for the QKV projections? Or a branch on `ndim == 3`.

scottjmaddox:
Rebased on the latest record: https://x.com/kellerjordan0/status/1866923518769893467; collecting runs

scottjmaddox:
Hmm, primeintellect's 8xH100 isn't as fast as @YouJiacheng, so even though this branch is still ~1ms faster per step, these runs aren't beating his 3.80s record...

logs-2024-12-10-record:
```
step:1480/1480 val_loss:3.2787 train_time:230396ms step_avg:156.73ms
```
230396ms = 3.84s

per-head-grad-orthonorm
```
step:1480/1480 val_loss:3.2799 train_time:228874ms step_avg:155.70ms
step:1480/1480 val_loss:3.2838 train_time:228884ms step_avg:155.70ms
```
228884ms = 3.81s

scottjmaddox:
Rebased back on master, to keep the diff clean. I'll have to see if I can find a faster 8xH100 to run on... Or perhaps the rules need some adaptation, to handle cases like this?

YouJiacheng:
looks good!

YouJiacheng:
one change need to be made: I group different matrices by size (numel) (and assume equal numel's matrices has ~equal time in Muon), maybe you need to group them by full shape.

YouJiacheng:
BTW, I think a better "method" to implement this is to add an extra attribute (maybe) called `._muon_shape` to the weight and check this attribute in the Muon optimizer.
```
self.c_q.weight._muon_shape = ...
```
so that the forward remains unchanged.

YouJiacheng:
Maybe you can add an extra tag to matrix parameters for grouping, see abov.

scottjmaddox:
Yeah, I'm seeing consistently slightly worse val_loss, now ☹️. Previous runs that didn't really show any degradation must have just been lucky.

per-head-grad-orthonorm:
```
e0e62795-b3ad-41bc-a9f5-bd563418d1fc
step:1480/1480 val_loss:3.2810 train_time:238423ms step_avg:162.19ms

359752c4-1166-4451-9017-2ad17427cc25
step:1480/1480 val_loss:3.2847 train_time:242806ms step_avg:165.17ms

be30f478-6e48-4b67-a51a-191999c615b0
step:1480/1480 val_loss:3.2814 train_time:247160ms step_avg:168.14ms
```

Running in docker on lambda labs 8xH100, I'm also getting periodic hangs that are killing the speedrun time 😢 
```
step:126/1480 train_time:17033ms step_avg:146.83ms
step:127/1480 train_time:17471ms step_avg:149.33ms
step:128/1480 train_time:18172ms step_avg:154.00ms
step:129/1480 train_time:18508ms step_avg:155.53ms
```

YouJiacheng:
uh, maybe you can try only split Q, K but not V?

scottjmaddox:
> uh, maybe you can try only split Q, K but not V?

Yeah, maybe. I had tried a few variations on 1xH100, and this one seemed like the best, but probably need more full runs to really be able to tell.

I'll let someone who isn't GPU poor take over from here 😅

leloykun:
I've already tried it

sharding only the QK weights is better than sharding QKV, but it's still worse that not sharding them

image

lapp0:
> BTW, I think a "better" method to implement this is to add an extra attribute (maybe) called `._muon_shape` to the weight and check this attribute in the Muon optimizer.
> 
> ```python
> class CastedLinear(nn.Linear):
>     def __init__(self, in_features, out_features, muon_shape: tuple[int] | None = None, muon_group: str | None = None):
>         super().__init__(in_features, out_features, bias=False)
>         self.weight._muon_shape = (out_features, in_features)  if muon_shape is None else muon_shape
>         if muon_group is not None:
>             self.weight._muon_group = muon_group
> 
>     def forward(self, x):
>         return F.linear(x, self.weight.type_as(x))
> 
> 
> class CausalSelfAttention(nn.Module):
>     def __init__(self, dim, num_heads):
>         super().__init__()
>         assert dim % num_heads == 0
>         self.num_heads = num_heads
>         head_dim = dim // num_heads
>         self.c_q = CastedLinear(dim, dim, muon_shape=(num_heads, head_dim, dim), muon_group="QK")
>         self.c_k = CastedLinear(dim, dim, muon_shape=(num_heads, head_dim, dim), muon_group="QK")
>         self.c_v = CastedLinear(dim, dim, muon_shape=(num_heads, head_dim, dim), muon_group="VO")
>         self.lambdas = nn.Parameter(torch.tensor([0.5, 0.5]))
>         self.rotary = Rotary(head_dim)
>         self.c_proj = CastedLinear(dim, dim, muon_group="VO")
>         self.c_proj.weight.data.zero_() # zero init suggested by @Grad62304977
> ```
> 
> so that the forward remains unchanged. and this allows us to explore more muon-orthonormalization-sharding strategies easily. (we have done some, but the results are not very good -- val_loss are worse)

Could you please help me with the intuition on why v_proj should be orthogonalized with c_proj and q_proj should be orthogonalized with v_proj? Thanks!

YouJiacheng:
oh, by "muon_group" I just mean the group in distributed muon, for work partition across different process (size of each group should be multiple of 8).

PR #56, opened 2024-12-19
Update Dockerfile to CUDA 12.6

bkoszarsky:
Speedup of 30 ms / step on 1xH100 -> ~5s total on 8xH100 to align between @YouJiacheng record and replication runs


YouJiacheng:
I can observe a ~0.5s speedup upon my environment. So I think we can get aligned results!!!
image

tysam-code:
Bumping, this seems useful to merge!

PR #59, opened 2024-12-20
WIP: ESM Baseline

lapp0:
Done:
- Update model parameters to reflect ESM++ vocab_size https://huggingface.co/Synthyra/ESMplusplus_large/blob/main/modeling_esm_plusplus.py#L868-L874
- Remove causal mask

Questions:
- What is the max sequence length we should consider when validating? 2048?
- Other than validation loss, what eval metrics do you recommend?

Remaining:
- validate with dataset
- 

When you're ready with the dataloader, I can merge it into my branch. Happy to help on the data side.


lapp0:
Wrong repo

PR #63, opened 2024-12-29
Add command line options

leloykun:
This PR allows usage of command line options. This should make it easier to do hyperparameter tuning.


KellerJordan:
adds too much complexity to add to the main file. but I'll leave the PR open in case people wanna see the code as an example.

PR #67, opened 2025-01-05
Remove redundancy rank assigning

LinglongQian:
Remove the redundancy of assigning `rank` and `self.rank` from the same environment variable.

### DIFF
--- a/train_gpt.py
+++ b/train_gpt.py
@@ -76,7 +76,6 @@ class Muon(torch.optim.Optimizer):
     """
     def __init__(self, params, lr=0.02, momentum=0.95, nesterov=True, ns_steps=5):
         self.world_size = int(os.environ['WORLD_SIZE'])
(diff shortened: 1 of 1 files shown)


KellerJordan:
I'm not a fan of global variables, sorry

LinglongQian:
Of course, I completely agree! Seeing that both `self.rank` and `rank` are mixed in `Muon`, I initially thought it was done to simplify the code as much as possible. In that case, it would be better to change `rank` in `Muon` to `self.rank`:
https://github.com/KellerJordan/modded-nanogpt/blob/89c97cea5ae51c7e9ef9b5d1d3bb25c32a4bbca1/train_gpt.py#L112
https://github.com/KellerJordan/modded-nanogpt/blob/89c97cea5ae51c7e9ef9b5d1d3bb25c32a4bbca1/train_gpt.py#L124

PR #68, opened 2025-01-09
Some (very minor) micro-optimizations (for fun)

awgu:
### Disclaimer
- **This is not meant as a real submission.** I had some time this morning and wanted to check out the current record. If anyone finds something interesting from this PR, then I am more than happy.
- This PR does only **numerics-preserving optimizations**, which naturally limits the impact since the current record is already well optimized. Overall, this PR gives very small improvement.
- I only have access to a power-capped 8xH100, which means my runs are slower. E.g., if I run the current record, I see 262191 ms = 4.37 min rather than 3.4 min. If I mention timing, I will do so with respect to my own setup.

### Overview
The PR includes a few different orthogonal changes. I did not have the time to ablate each one in real runs, but I will present them roughly in order of decreasing impact and increasing complexity.

The first thing I did is look at the profiler trace of the current record. From there, I see three (very minor) inefficiencies, which will correspond to the three optimizations in this PR:
1. Beginning of forward: There is a ~1.730 ms gap between the previous optimizer step and the first forward kernel, where there is only a small copy kernel and dataloading H2D kernels. 

Screenshot 2025-01-09 at 11 00 13 AM

This is mainly due to the `cudaDeviceSynchronize` from the H2D copy for the flex attention sliding window number of blocks:
https://github.com/KellerJordan/modded-nanogpt/blob/e48427208f14a0a27ef0fcf2d369a0f4299342d7/train_gpt.py#L499
**Optimization 1: Replace `.copy_()` with `.fill_()`.**
Under the hood, when we call `device_scalar_tensor.copy_(python_float)`, PyTorch will issue a H2D copy followed by `cudaDeviceSynchronize`, blocking the CPU until that copy finishes. We can equivalently use `fill_()` that directly launches a kernel to write the scalar value to the device tensor. This means that the CPU can run ahead and hides CPU overheads like the DDP pre-forward and Dynamo cache lookup.

2. Muon optimizer: The current implementation uses `dist.all_gather()`, which includes device copies for outputs in addition to the NCCL all-gather call. Because the data size being all-gathered is not large and these copies can mostly overlap with optimizer compute, this only contributes ~0.2 ms extra overhead.
Screenshot 2025-01-09 at 11 11 56 AM

**Optimization 2: Use `dist.all_gather_into_tensor()`, not `dist.all_gather()`, to avoid extra copies.**
We should instead use `dist.all_gather_into_tensor()` to avoid the device copies and only call NCCL all-gather. To do this, we simply construct a flat tensor for the all-gather output and cache views into it instead of having per-rank all-gather outputs that are not together contiguous in memory. This is still a relatively contained change in the `Muon` implementation.

3. End of backward: There is a ~2.3 ms gap at the end of backward from the final exposed gradient all-reduces from DDP. Is it possible to make this gap even smaller?

Screenshot 2025-01-09 at 11 15 15 AM

One idea is to use some optimizer computation to fill in part of that gap since some parameters' gradients have already been all-reduced and can start their optimizer step. _While the idea is simple, my implementation is involved and almost surely should not be included into future records due to the complexity._

**Optimization 3: Reorder some optimizer computation with final all-reduce for better overlap.**
I chose to isolate the embedding parameters to try to overlap their gradient all-reduce with other optimizer compute. To do so:
- Ignored them from DDP, choosing to manage their gradient all-reduce myself
- Separated them from `optimizer1` into a new Adam optimizer instance
- Implemented a batched all-reduce (foreach copy -> async all-reduce, where the copy is to copy into a "bucket")
- Registered a special backward hook to run this all-reduce (notably, before the end of backward) and did so using a separate process group so that the NCCL all-reduce runs in a separate CUDA stream
- Reordered the optimizer computation accordingly

To give some more details on some nuances:
- We want to run the embedding gradient all-reduce before the end of backward so that we can overlap the foreach copy with the last DDP all-reduce. DDP will wait for all of its all-reduces at the very end of backward, so as long as we can run the foreach copy before the end, we can overlap. We can only start this foreach copy once all participating gradients have been computing (i.e. those of the embedding parameters). PyTorch has an API `register_post_accumulate_grad_hook` that runs a hook when _one_ parameter's gradient has been accumulated. https://github.com/pytorch/pytorch/pull/131949/ generalizes this to an "all" hook that only runs once all passed-in tensors' gradients have been accumulated, but we held off on landing this PR since there was not broad applicability. So, all we are doing is saying, once _all_ embedding parameters' gradients have been computed, run a hook -- this hook copies the gradients into a "bucket" and issues an async all-reduce that will be waited on later in the optimizer step.

Example: embedding gradient foreach copy overlaps with last DDP all-reduce
Screenshot 2025-01-09 at 11 31 38 AM

- Every NCCL process group has its own internal stream (per device). This means that, since we are only using the default process group over all workers, if even if we issue an async all-reduce for the gradients, the subsequent Muon all-gather will serialize after the all-reduce. This generally makes sense: NCCL collectives use SMs and network resources and overlapping collectives over the same ranks often hurts net performance. However, in our case, we find a small net win by overlapping the embedding gradient all-reduce, Muon all-gather, and Muon optimizer compute (possibly because the memory-bound optimizer is not too harshly affected by the SM contention).

Example: Muon all-gather serializes after embedding all-reduce when using the same default process group for both
Screenshot 2025-01-09 at 11 29 11 AM

Example: using a separate process group for the all-reduce allows it to run in a separate stream -- this reduces the exposed gap from ~1.5 ms to ~0.9 ms
Screenshot 2025-01-09 at 11 30 47 AM

**Other misc. changes:**
- Issued next batch's H2D in the current step's optimizer -- not too important after the `copy_` -> `fill_` change
- Disabled Python `gc` to minimize opportunity for variation across ranks -- did not see much difference; interestingly, my avg step time increases with the # of steps, but I have not investigated why

---

In summary: Locally, my step time is ~173 ms. Optimization 1 saves ~1.7 ms, Optimization 2 saves ~0.2 ms, and Optimization 3 saves ~1 ms. 

maybe cc: @YouJiacheng as someone who might be even partially interested by these changes


YouJiacheng:
I considered Optimization 3 (cuz embedding is large and we can run other optimizers first) but I thought excluding them from DDP then manual all reduce is too ugly.
But didn't know it can reduce 1ms step time… this is quite significant.

YouJiacheng:
For optimization 1, my original code only call that copy_ when the value changes, so I only call it 14 times. (0.017ms on average)

but thank you for identifying this cause of perf discrepancy!

YouJiacheng:
btw I'm curious why the optimizer.step can't start before the all reduce finish?

If I set lm_head as the first parameters group (each paramters group will run a separate fused_adam kernel), can it run before the all reduce finish?

IIUC, lm_head.weight.grad should be ready at the end of backward?

awgu:
@YouJiacheng I thought `lm_head.weight` is used at the end of forward, so its grad is ready at the beginning of backward. I am overlapping the optimizer step of `lm_head.weight` and 1D parameters with the embedding gradient all-reduce.

YouJiacheng:
image
here is the original implementation.
btw, IIUC, set `non_blocking=True` will only block host (CPU) until the data are copied into a pinned memory? (not sure)

YouJiacheng:
> so its grad is ready at the beginning of backward. 

yeah I meant it should be already ready at the end of backward because its bucket's all reduce starts first.

YouJiacheng:
> I am overlapping the optimizer step of `lm_head.weight` and 1D parameters with the embedding gradient all-reduce.

my question is, why this can't be achieved by simply run the optimizer step of `lm_head.weight` first?
there are some fake dependencies?

awgu:
@YouJiacheng I think the issue with DDP is that it will wait on all of the all-reduces before the `.backward()` returns back to the user. This means subsequent compute queued on the default/compute stream will be serialized after the all-reduces, e.g. the optimizer compute, even if there is no true data dependency. I am not sure if this is what you were asking about.

---

Regarding H2D `device_tensor.copy_(python_float, non_blocking=True)`: I think this avoids the `cudaDeviceSynchronize` as well and works equivalently:
Screenshot 2025-01-09 at 2 01 51 PM
(so your original impl was good!)

YouJiacheng:
thank you!
but theoretically `fill_` is faster because there isn't a H2D copy from pageable memory (not sure how passed-by-value arguments are passed to a GPU kernel but I guess the cost is included in the kernel launch cost).

YouJiacheng:
I came up with a workaround for Optimization 3, not sure if it can work:
1. `register_post_accumulate_grad_hook` for `lm_head.weight`, record an event.
2. run the optimizer step of `lm_head.weight` in another stream, wait for that event.

YouJiacheng:
I didn't know `torch.distributed.all_gather` isn't in-place (requires extra copies).
So... there is only a NCCL AllGather kernel that works for a single tensor (continuous memory) as target?

awgu:
> I came up with a workaround for Optimization 3, not sure if it can work:
> 
> 1. `register_post_accumulate_grad_hook` for `lm_head.weight`, record an event.
> 2. run the optimizer step of `lm_head.weight` in another stream, wait for that event.


@YouJiacheng off the top of my head, I have some concern about step 1.

In `ProcessGroupNCCL`, when you call something like `dist.all_reduce()`, the logic is like:
- Record event in current stream
- Process group stream waits for current stream
- Run NCCL kernel in process group stream and record event
- Return work object

Calling `work.wait()` has current stream wait for the recorded event in the process group stream. If `async_op=False`, then the Python layer calls `work.wait()` before returning back to the user (e.g. calls it in `dist.all_reduce()`).

The issue is that you need to record an event after the NCCL all-reduce kernel, but the process group stream is not exposed today in PyTorch. 

So you need to take some care here. I am not entirely clear if the `register_post_accumulate_grad_hook` works with DDP and if so how.
1. Suppose it runs after DDP all-reduce but before the final wait on all-reduce that runs at the end of backward. If so, you still have no way to record an event after the NCCL all-reduce itself since that process group stream is not exposed and the default stream has no dependency yet with the all-reduce (since no `work.wait()` call).
2. Suppose it runs before DDP all-reduce. Then it is definitely not going to be right.


---

> So... there is only a NCCL AllGather kernel that works for a single tensor (continuous memory) as target?

Yes, I think so.

YouJiacheng:
> the issue with DDP is that it will wait on all of the all-reduces before the `.backward()` returns back to the user.

can you give a pointer/link to material/source code for this? does it call cuda synchronize or just wait for all handles?

awgu:
One trick (which we use in FSDP implementation in one code path) is that we introduce a dummy stream to serve as the current stream when issuing a c10d collective. That way, when the c10d collective has the process group stream wait for the current stream, that current stream is _not_ the default stream but rather some separate stream we introduced and can sync against. 

The issue is that I do not think we can customize DDP to run its all-reduce in a separate stream. Maybe with the comm hook... 😆 

---

I am not a DDP expert, but I _think- the code pointer for the waiting is here:
https://github.com/pytorch/pytorch/blob/6bc17b0725f8adc1b7293dd44c90e8a6c495ea03/torch/csrc/distributed/c10d/reducer.cpp#L1620

train_gpt.py
```diff
@@ -93,7 +158,12 @@ def step(self):
             momentum = group['momentum']
             nesterov = group['nesterov']
             ns_steps = group['ns_steps']
-            update_buffers = group['update_buffer']
+            update_buffer = group['update_buffer']
+            if "update_buffer_views" not in group:
+                group["update_buffer_views"] = [
+                    update_buffer[i].view(-1) for i in range(self.world_size)
+                ]
```
YouJiacheng, line 165: This in `__init__` looks more compact (the codes for `update_buffer` and its views are put together)
```python
        def create_update_buffer(size: int):
            b = torch.empty(world_size, size, dtype=torch.bfloat16, device="cuda")
            return dict(update_buffer=b, update_buffer_views=[b[i] for i in range(world_size)])
        param_groups = [
            dict(params=[p for p in params if p.numel() == size], **create_update_buffer(size)) for size in sizes]
```

YouJiacheng:
> I do not think we can customize DDP to run its all-reduce in a separate stream.

It doesn't use current stream (set by with stream) but the global default stream?

awgu:
> > I do not think we can customize DDP to run its all-reduce in a separate stream.
> 
> It doesn't use current stream (set by with stream) but the global default stream?

should be the current stream -- I just meant that it might be hard to customize which is the current stream when the all-reduce runs, but maybe with the [communication hook](https://pytorch.org/docs/stable/ddp_comm_hooks.html), it is possible

YouJiacheng:
> So you need to take some care here. I am not entirely clear if the `register_post_accumulate_grad_hook` works with DDP and if so how.

Uh you're right, the interaction between this hook and DDP is unclear.

YouJiacheng:
maybe we can just exclude `lm_head` from DDP, and then use `register_post_accumulate_grad_hook` to run all reduce & optimizer step for it? the concern is the optimizer step might be launched too early so that it can't fill the gap.

YouJiacheng:
btw, is it safe to have aliasing (memory address ranges overlap, i.e. source is a view or target) in all_gather_into_tensor's source and target (safe = correct & no performance issue)?

awgu:
> btw, is it safe to have aliasing (memory address ranges overlap, i.e. source is a view of target) in all_gather_into_tensor's source and target (safe = correct & no performance issue)?
> 
> oh there is a special handling in our case? 

If the source is the correct view into the target (like the corresponding offset for that rank), then NCCL should dispatch into an in-place all-gather, which is every-so-slightly faster since it avoids re-copying the data for the source. It does not make much difference at larger scales though, maybe some small difference at smaller scale.

YouJiacheng:
Cuz Muon is a distributed optimizer, I have a plan to write a custom DDP using DTensor and perform reduce-scatter instead of all-reduce. But I have a few concerns:
1. Muon is a matrix optimizer, so we can't shard a matrix -- instead, a group of matrices must be copied into a tensor to perform reduce-scatter, I wonder if this copy can be avoided? Maybe the implementation of `gradient_as_bucket_view` in DDP is relevant.
2. Does PyTorch have some auto-bucketization mechanism for DTensor's communication? Or I have to do it by myself.

YouJiacheng:
Wow I just found that you are the FSDP2 & DTensor expert! https://github.com/pytorch/pytorch/issues/114299
Do you have any material about how to implement a DDP (with good performance) using DTensor?

awgu:
@YouJiacheng I might be a bit flaky in the discussion -- I apologize in advance, and I will try to check when I can and think more too.

My personal opinion is for this kind of work, you should not use DTensor. DTensor is an abstraction trying to hide some communication details from users (such that they only need to think about what the sharding change is, e.g. from `Shard(0)` to `Replicate()`). 

However, in this case, we have a very precise goal and need full manual control. It will be easier to implement yourself rather than try to use DTensor and then verify that it is doing what you expect. Let me think more about the data layout for Muon's sharding later (to be honest, I did not read the code too closely yet 😅 ).

awgu:
Maybe I can make an attempt at using reduce-scatter instead of all-reduce some time soon (maybe this weekend). One thing on my mind is that I think it would need to be a custom solution tailored to the model arch with some constraints. Otherwise, preserving the matrix sharding is hard. Let me know if you think of anything better 😃

YouJiacheng:
full manual control is definitely better for performance, but it might increase the complexity too much, so I hope DTensor can be a good tradeoff.

awgu:
closing this since @YouJiacheng has finished taking a look

PR #71, opened 2025-01-16
Long-Short Sliding Window Attention (3.2 sec or 0.053 mins improvement)

leloykun:
Currently, we warmup the context length of the sliding window attention at the same rate in all layers. This attempt warms up the context length differently in some layers instead. This leads to a ~3 ms/step improvement. However, to compensate for the increase in `val_loss`, we needed to add 15 more training steps. Thus, overall, this saves ~3.2 secs on our 8xH100 pod.

- c @YouJiacheng for optimizing and simplifying the code for the sliding window block attention. His efforts made implementing this change much easier.

image
image

---

With the updated version of PyTorch (nightly v. 25/01/10), the wallclock runtime of the baseline is now just `202898 secs`. This runs at `199699 secs` so ~3.2 secs improvement


leloykun:
Closing as the PR for the 2.99 min record (which employs this trick) has already been merged.

PR #74, opened 2025-01-21
[WIP] BFloat16 weights; 2 sec improvement

leloykun:
From the pytorch trace below, see that the current record is bottlenecked by the backwards pass. Thus, it makes sense to shift our focus on optimizing this part of the run.

Converting the weights to bfloat16 is the lowest hanging fruit we can do & it seems to reduce wallclock time by 2 secs.

image

### DIFF
--- a/train_gpt.py
+++ b/train_gpt.py
@@ -470,7 +470,7 @@ class Hyperparameters:
     # optimization
     batch_size = 8*64*1024 # batch size in tokens
     num_iterations = 1400 # number of iterations to run
-    cooldown_frac = 0.4 # fraction of training spent cooling down the learning rate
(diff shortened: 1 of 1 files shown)


YouJiacheng:
I think a better choice is to also implement a master weight in optimizers so we can get back to FP32 with minimal performance regression when we need FP32.
Note that embeddings are already full BF16, so we only need to deal with `lm_head` and weights updated by Muon.
BF16 accumulation is safe now (probably) because we only decay the lr to 0.1× peak after previous changes.
But in longer runs (ofc, irrelevant to) we might decay the lr to smaller values.

leloykun:
closing as I have a better PR https://github.com/KellerJordan/modded-nanogpt/pull/77

PR #77, opened 2025-01-26
2.923 min. record

leloykun:
## Changelogs:

1. Reduced per-device training sequence length from `64*1024` to `48*1024`. See [Critical Batch Size](https://arxiv.org/abs/2410.21676) literature.
2. Increased per-device eval sequence length from `64*1024` to `4*64*1024`. This improves `val_loss` by `~0.0015` or an equivalent of a reduction of 10 training steps. Overall it saves `~1 sec` of training time.
3. Modified scales for `fp8` training of LM Head. Saves `1 sec` and improves `val_loss` by as much as `~0.01` after reducing training sequence length down to `48*1024`. I don't know wtf is causing this and I'm NOT going crazy about this. I have evidence. See `records/012625_MiscTweaks/no-autocast-same-fp8-scales`.
    - `w_s = 2.0**9` (from `2.0**5`)
    - `grad_s = 2.0**19` (from `2.0**29`)
4. Upgraded PyTorch to 2.7.0 nightly version (20250125) for CUDA 12.6
     - `pip install --pre torch==2.7.0.dev20250125+cu126 --index-url https://download.pytorch.org/whl/nightly/cu126`

Note: effects of (3) were previously misattributed to the autocast

---

## Evidence for record

image
image

```python
accs = [3.2806, 3.2771, 3.2829, 3.2813, 3.2789, 3.2774, 3.2798, 3.2759, 3.2794, 3.2775, 3.2768, 3.2793, 3.2838, 3.2779, 3.2782, 3.277, 3.2775, 3.2784, 3.2782, 3.2776, 3.2814, 3.2785, 3.2793, 3.2797, 3.2782, 3.2789, 3.2759, 3.2803, 3.278, 3.2782]

import scipy.stats
print('p=0.00' % scipy.stats.ttest_1samp(accs, 3.28, alternative='less').pvalue)
# p=0.0007 (statistically significant)

import torch
print(torch.std_mean(torch.tensor(accs)))
# (tensor(0.0019), tensor(3.2788))
```

---

## Ablation Results

- With the old training sequence length (64*1024), the new fp8 scales slightly improves `val_loss` but also slightly worsens wallclock time
- With the new training sequence length (48*1024), the new fp8 scales significantly improves `val_loss` (~0.01) _and_ also slightly improves wallclock time (~1 sec)
- Quadrupling eval sequence length improves `val_loss` by a `~0.0015`--which allows us reduce training steps by 10

image


YouJiacheng:
> Calculate backward passes of linear layers in `bfloat16`, but keep the weights in `float32`

`autocast` can do this? Not sure if that's the exact behavior.

leloykun:
> > Calculate backward passes of linear layers in `bfloat16`, but keep the weights in `float32`
> 
> `autocast` can do this? Not sure if that's the exact behavior. Previous implementation already do backward in `bfloat16`, but it might have some extra costs that are not optimized by torch compile yet.

it says so in their docs: https://pytorch.org/docs/stable/notes/amp_examples.html

image

are there implementation details I'm missing?

leloykun:
hmmm, you may be right... this may actually only be helping the compiler optimize stuff

train_gpt.py
```diff
@@ -296,36 +301,38 @@ def __init__(self, dim: int):
...
+        self.num_layers = num_layers
+        self.num_embeddings = num_embeddings
+        self.embed = nn.ModuleList([nn.Embedding(vocab_size, embedding_dim) for _ in range(num_embeddings)])
 
-    def forward(self, input_seq) -> list[Tensor | None]:
+    def forward(self, input_seq: Tensor) -> list[Tensor | None]:
         ve = [emb(input_seq) for emb in self.embed]
         # 012 ... 012 structure on token value embeddings by @YouJiacheng, improved on @leloykun's U-net structure
-        ve = [ve[0], ve[1], ve[2], None, None, None, None, None, None, ve[0], ve[1], ve[2]]
+        ve = [ve[0], ve[1], ve[2]] + [None] * (self.num_layers - 2 * self.num_embeddings) + [ve[0], ve[1], ve[2]]
```
linux-leo, line 335: that just seems a little overcomplicated, if you change the number of embeddings, it won't work anyway without changing the array. so I think having it hardcoded gives more flexibility.

leloykun: tbh, I prefer this cuz it's one less thing to manually modify when scaling up to the medium-size records

PR #80, opened 2025-02-12
2.941 max-autotune record

notlober:
Hi, just learned about this benchmark and wanted to contribute a pr from my experiences, so i just added mode="max-autotune-no-cudagraphs" to torch.compile.

3.5 seconds faster than [020125_RuleTweak](https://github.com/KellerJordan/modded-nanogpt/tree/master/records/020125_RuleTweak)

train_time:176467ms vs train_time:179828ms and step_avgs are faster now

also tried "max-autotune" too but error occurs, compilation time did not change much so i think its valid.

trying to make "max-autotune" and some other args like "epilogue_fusion" of torch.compile can help, it was all i could test with 20 usd spent on runpod

### DIFF
--- a/records/021225_MaxAutotune/11e9d28c-fde3-40f6-adaf-398a49a14b93.txt
+++ b/records/021225_MaxAutotune/11e9d28c-fde3-40f6-adaf-398a49a14b93.txt
@@ -1,0 +1,2479 @@
+import os
+import sys
+with open(sys.argv[0]) as f:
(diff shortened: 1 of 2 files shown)


KellerJordan:
sorry, extra torch.compile options are banned. they cause the run to take way too long to compile. it's supposed to be part of rule 3, but I see it isn't totally clear now. I'll make it more clear

PR #87, opened 2025-03-05
add deepseek style pos embedding

notlober:
this pr adds deepseek style pos embedding to attention layers, now loss is better overall.

simply applies rotary embedding to half of query and keys (:head_dim//2) and keeps rest of query and key with no pos embedding (head_dim//2:)

originally deepseek also adds low rank compressions to reduce kv cache but in this training setup its not needed so just no pos+rope added.

tested with 8xh100


KellerJordan:
I believe the current code already uses half no-op RoPE

would be happy to accept this change if you can show it's a new record; closing until then

notlober:
https://drive.google.com/file/d/1f8D4wrXJnSIS4U862aztVWi6q7Sk0Tkh/view?usp=sharing

last val loss: 3.2773
last record original repo:3.2788

PR #89, opened 2025-03-13
Add efficient validation on HellaSwag

trianxy:
This PR adds optional validation using HellaSwag accuracy (on 10042 validation examples).

It is controlled by setting the hyperparameter `hellaswag=True`.

*Motivation*
I am trying to make the model smarter by adjusting the architecture/training objective. And I wanted to have benchmarks like HellaSwag which may better correlate with smarter models than loss. By the way: Having this repo which makes a model converge quickly is super helpful!

*Not sure if it's worth it to merge into master. Some thoughts:*
- Pros:
  - more data/confidence on whether a change improved the model
  - related to the above bullet point, it can improve comparability to other models and training runs
- Cons:
  - adds ~150 lines of code
  - adds ~40s wall time to script when running on 8 x H100:
    - ~30s to download data + preprocess + warm-up
    - ~6s to run validations (0.4s every time we run validation on 10042 examples: the code packs 400-800 HellaSwag tasks into 1 sequence which is then evaluated using 1 forward pass. 2 forward passes are required for each of the 8 GPUs to process all examples)

Happy to adjust the PR if helpful.

### DIFF
--- a/evals/hellaswag.py
+++ b/evals/hellaswag.py
@@ -1,0 +1,194 @@
+from typing import Callable
+import time
+from dataclasses import dataclass
+from functools import lru_cache
+
+import torch
+from torch import Tensor
+import torch.distributed as dist
+from datasets import load_dataset
+import tiktoken
+from huggingface_hub import logging as hf_logging
+hf_logging.set_verbosity_error()
+
+tokenizer = tiktoken.get_encoding("gpt2")
+rank = dist.get_rank()
+world_size = dist.get_world_size()
+EOT_ID = tokenizer.eot_token
+
+@dataclass
+class HellaswagTask:
+    # store indices (start, end_excluding) of 4 endings with respect to the targets in PackedHellaswagSequence
+    spans: tuple[tuple[int,int], tuple[int,int], tuple[int,int], tuple[int,int]]
+
+    # label of correct ending of task
+    label: int
+
+
+@dataclass
+class PackedHellaswagSequence:
+    """Contains multiple HellaSwag tasks in 1 token sequence for efficient evaluation.
+
+    If seq_len is ~260k we pack ~800 tasks into 1 sequence. For each task we store the
+    tokens of 4 strings with each string being the context plus a possible ending.
+
+    self.inputs does not store the last token of each ending because it's not used for 
+    evaluation. Similarly, self.targets does not store the first token of each context.
+    """
+                               # EXAMPLE:
+                               # The below sequence starts with Task0, which has context [10,11] 
+                               # and 4 endings [20,21],[30,31],[40](correct ending),[50].
+                               #            -Task0--------------------------------------- -Task1...
+                               # indices     0  1  2   3  4  5  6   7  8  9  10 11 12  13 14 ...
+    inputs: Tensor             # inputs  = [10 11 20 EOT 10 11 30 EOT 10 11 EOT 10 11 EOT 24 ...
+    targets: Tensor            # targets = [11 20 21 EOT 11 30 31 EOT 11 40 EOT 11 50 EOT 37 ...
+    tasks: list[HellaswagTask] # .spans        [1, 3),      [5, 7),      [9,10),  [12,13)    ...
+                               # .label                                   2                  ...
+    doc_end_positions: Tensor
+
+
+def finalize_sequence(inputs: list[int], targets: list[int], tasks: list[HellaswagTask], seq_len: int) -> PackedHellaswagSequence:
+    doc_len_min_average = 70 # use a smaller number than the median doc length of 94 (doc=context+one ending)
+    max_num_docs = ((seq_len // doc_len_min_average) // 128 + 1) * 128  
+    pad_id = 0  # can use any token id except EOT_ID which is reserved to delimit HellaSwag tasks and endings
+
+    # pad to seq_len
+    padding_length = seq_len - len(inputs)
+    inputs.extend([pad_id] * padding_length)
+    targets.extend([pad_id] * padding_length)
+    assert len(inputs) == len(targets) == seq_len
+
+    inputs_tensor = torch.tensor(inputs, dtype=torch.int32)
+    targets_tensor = torch.tensor(targets, dtype=torch.int64)
+
+    cum_lengths = torch.nonzero(inputs_tensor == EOT_ID)[:, 0] + 1  # add +1 to assign EOT_ID to preceding doc
+    _cum_lengths = torch.full((max_num_docs,), seq_len)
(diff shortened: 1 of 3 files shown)


gszfwsb (approved):

trianxy:
Hey @gszfwsb , thanks for the approval. Before we merge, let me implement what @KellerJordan mentioned in https://github.com/KellerJordan/modded-nanogpt/pull/89#issuecomment-2908125388 and also resolve the conflicts.

Haven't gotten around it yet, but am looking to do it in September.

After that, I'd ask you to take one more look again.

trianxy:
Hey @gszfwsb , I just updated this PR to incorporate @KellerJordan 's suggestion to move the HellaSwag code into a separate file `data/hellaswag.py`.

To both of you: Feel free to let me know if there's anything else I can do on my side.

FYI
---
I didn't change how `train_gpt.py` works, but I kind of had to move a bunch of code behind a clause `if __name__ == "__main__"...`, because I wanted to import the model from that file without re-running the training loop.

There would have been a few options to implement this PR without touching `train_gpt.py`, e.g.
- Option 1: read the source code of `train_gpt.py` into a string, remove unnecessary stuff, and then `exec(...)` it
- Option 2: use the module `ast` to only load from the source code those objects which are necessary to define the model architecture

I decided against these, because they felt less maintainable and more likely to break in the future.

ClassicLarry:
Does this work if we only merge in the hellaswag.py file? Maybe we need some updates to enable the YaRN parameters to be saved in the model.torch file. I am interested to get something of this sort merged in, to enable the occasional sanity check that the model is not data snooping and general evals look good. The model is starting to reach a stage where it can be complex to reason about cross token interactions.

trianxy:
> Does this work if we only merge in the hellaswag.py file?...

No. I think we need a small change to `train_gpt.py` as well:

- `hellaswag.py` needs to run `model = GPT(...)` (e.g. via `from train_gpt import model`) and then load a checkpoint into it. 
- This requires moving some of the code in `train_gpt.py` behind a `if __name__ == '__main__'` flag to avoid starting a training job when doing the import

I can try to implement this with as few code line changes as possible. What do you think @ClassicLarry ?

Open to other suggestions as well.

ClassicLarry:
> > Does this work if we only merge in the hellaswag.py file?...
> 
> No. I think we need a small change to `train_gpt.py` as well:
> 
> * `hellaswag.py` needs to run `model = GPT(...)` (e.g. via `from train_gpt import model`) and then load a checkpoint into it.
> * This requires moving some of the code in `train_gpt.py` behind a `if __name__ == '__main__'` flag to avoid starting a training job when doing the import
> 
> I can try to implement this with as few code line changes as possible. What do you think @ClassicLarry ?
> 
> Open to other suggestions as well.

Sounds good to me. train.py would include __main__ check, maybe one update to model.forward() to return logits if no target, and might need to recheck the model logging for YaRN parameters or things like that. (would want to confirm that export/import works and gives 3.28 loss)

imo the __main__ indentation is slight hit to readability, but scenario where bunch of PRs are getting merged with data leakage would be far worse, and eval should help protect against that. I manually ran an eval after the bigram one but not going to keep manually running evals.

trianxy:
>imo the main indentation is slight hit to readability,

I don't like the main indentation either.

How about I implement it in a way that HellaSwag can be run during training by setting a hyperparameter `hellaswag=True`? (if that hyperparameter is False, then no imports or other stuff would happen to ensure there is no hit to wall-clock time) 

Most of the code would still live in a new `hellaswag.py`. This would make running HellaSwag easier, since you don't need to first save a checkpoint and then run it on that checkpoint.

This was my initial idea 1y ago, which we had discarded in order to keep the main code path simple. But the importance  of easily running an additional validation has maybe increased since then.

ClassicLarry:
> > imo the main indentation is slight hit to readability,
> 
> I don't like the main indentation either.
> 
> How about I implement it in a way that HellaSwag can be run during training by setting a hyperparameter `hellaswag=True`? (if that hyperparameter is False, then no imports or other stuff would happen to ensure there is no hit to wall-clock time)

sounds good. can have something simple like import run_hellaswag_eval, then on last line if master_process and args.run_evals: run_hellaswag_eval(model)

trianxy:
> sounds good. can have something simple like import run_hellaswag_eval, then on last line if master_process and args.run_evals: run_hellaswag_eval(model)

Alright, I'm aiming to update this PR accordingly until the end of this week.

trianxy:
Hey @ClassicLarry , this PR is now ready for review.

The default is to NOT impact the training run since `args.run_evals=False`.

If the latter hyperparameter is set to True, the accuracy on 10042 HellaSwag validation tasks will be calculated and printed at the end of the training run.

The first time on a fresh machine, this eval takes 15s. Afterwards 5s. If it’s run multiple times in the same Python process, the >= 2nd run will take 0.5s.

trianxy:
The only minor change I implemented to the training was having the model return the loss per token. This does not significantly change the loss or training time: On average, over 10 runs, the change yielded 69ms SLOWER training time and 0.001 BETTER validation loss, which sounds fine to me. More precisely:

- Current master commit e22a34b (1.496min, 02/10/26), 10 runs: mean validation loss `3.27923 (std 0.00128)` and mean training time `96327ms (std 342ms)`. The 8xH100 SXM machine I used is ~6s (!) slower than the current record. _Is this ok?_
- Same machine, same commit + loss_per_token change, 10 runs: mean validation loss `3.27827 (std 0.00086)` and mean training time `96396ms (std 235ms)`, **mean HellaSwag accuracy `29.976% (std 0.196%)`**


I also checked the HellaSwag accuracy of 2 past commits. Things look good: today’s HellaSwag accuracy of 29.976% is similar or better than:

- Commit 93f0e6b + current PyTorch (1.655min, 01/19/26) + this PR, 10 runs: accuracy `29.930% (std 0.244%)` (missed noting the loss and trainingtime..)
- Commit 587b45f + current PyTorch (2.997min, 02/01/25) + this PR, 10 runs: accuracy `29.4782% (std 0.1955%)`, mean validation loss `3.28135 (std 0.00176)`, mean trainingtime `177235ms (std 136ms)`

Finally, on the current master commit e22a34b, I checked what I get when increasing the steps from 1530 to:

- 4000 steps and 0.70 cooldown_frac (1 run): validation loss `3.2484`, trainingtime `248938ms`, HellaSwag accuracy `32.513%`
- 10000 steps and 0.95 cooldown_frac (1 run): validation loss `3.0338`, trainingtime `626025ms`, HellaSwag accuracy `34.963%`

ClassicLarry:
cool looks good. I'll test and merge later.

PR #97, opened 2025-04-25
Hide gradient communication without losing fusions

xmfan:
Not a submission, just wanted to give a heads up that with a fix in recent PyTorch nightly (https://github.com/pytorch/pytorch/pull/150074) and enabling Compiled Autograd should allow you to hide the DDP-style gradient allreduces with the backward compute, without losing too much on fusions, cc @YouJiacheng.

The delta is:
1. Capturing allreduces into the graph: This is crucial to be able to jointly schedule compute and communication without relying on graph breaks. Compiled Autograd lets you do this easily with backward hooks, but you could also do this by writing custom autograd functions.
2. Letting fusions happen across collectives: That PyTorch-side PR fixes an issue where collectives were delayed until the end of the backward. During inductor passes, there's quite a bit of movement that happen across the collective nodes, that couldn't have happened if there were graph breaks instead. This explains the poor performance of DDPOptimizer (overlap via graph-break) as it prevented many fusions from happening.

Based on my runs on some nerfed H100s, it should save you ~1-2ms per iteration without logic change. 

Before: 116.85ms per iteration (~7ms exposed)
image

After: 114.94ms per iteration (~1.7ms exposed)
image

### DIFF
--- a/train_gpt.py
+++ b/train_gpt.py
@@ -23,6 +23,11 @@ from torch.nn.attention.flex_attention import BlockMask, flex_attention
 # -----------------------------------------------------------------------------
 # Custom operators: FP8 matmul by @YouJiacheng
 
+def ca_ctx():
+    # Compiled Autograd's config API doesn't work for mm_backward_op's torch.compile call,
+    # directly use context manager instead. Tracked in https://github.com/pytorch/pytorch/issues/152219.
+    return torch._dynamo.compiled_autograd._enable(torch.compile)
+
 @torch.library.custom_op("nanogpt::mm", mutates_args=())
 def mm_op(x: Tensor, w: Tensor, x_s: float, w_s: float, grad_s: float) -> tuple[Tensor, Tensor, Tensor]:
     @torch.compile
@@ -544,6 +549,12 @@ def get_window_size_blocks(step: int):
     window_size = next_multiple_of_n(1728 * x, n=128)
     return get_window_size_blocks_helper(window_size)
 
+def ar_hook(param):
+    dist.all_reduce(param.grad, op=dist.ReduceOp.AVG)
+
+for param in model.parameters():
+    param.register_post_accumulate_grad_hook(ar_hook)
+
 model: nn.Module = torch.compile(model, dynamic=False)
 
 ########################################
@@ -556,9 +567,8 @@ initial_state = dict(model=copy.deepcopy(model.state_dict()),
                      optimizers=[copy.deepcopy(opt.state_dict()) for opt in optimizers]) # save the initial state
 for _ in range(warmup_steps):
     inputs = targets = torch.randint(0, args.vocab_size, size=(args.train_seq_len,), device="cuda")
-    model(inputs.to(torch.int32), targets, get_window_size_blocks(0)).backward()
-    for param in model.parameters():
-        dist.all_reduce(param.grad, op=dist.ReduceOp.AVG)
+    with ca_ctx():
+        model(inputs.to(torch.int32), targets, get_window_size_blocks(0)).backward()
     for opt in optimizers:
         opt.step()
     model.zero_grad(set_to_none=True)
@@ -615,9 +625,8 @@ for step in range(train_steps + 1):
 
     # --------------- TRAINING SECTION -----------------
     inputs, targets = next(train_loader)
-    model(inputs, targets, get_window_size_blocks(step)).backward()
-    for param in model.parameters():
-        dist.all_reduce(param.grad, op=dist.ReduceOp.AVG)
+    with ca_ctx():
+        model(inputs, targets, get_window_size_blocks(step)).backward()
     # set optimization hyperparameters
     for opt in optimizers:
         for group in opt.param_groups:


YouJiacheng:
Yep I have been considering this for a long time but I'm not sure about the state of compiled autograd + flex attention.
https://github.com/pytorch/pytorch/issues/142313
https://github.com/pytorch/pytorch/issues/143180
https://github.com/pytorch/pytorch/pull/144533
It seems to be fixed after 20250117? (So I recently added back the `compiled_autograd` but haven't done anything yet)
https://github.com/KellerJordan/modded-nanogpt/blob/a202a3a0ca99d69bb7f847e5337c7c6e0890fd92/train_gptm.py#L21


Note: it's hard to move version beyond nightly 20250208 because of an unexpected performance regression: https://github.com/pytorch/pytorch/issues/147463 (even worse, there seems to be further regression after that...)

YouJiacheng:
If that needs a version beyond `20250208`, then I need to investigate what causes the performance regression by checking the inductor output etc. 🥹.

YouJiacheng:
BTW, based on my naive understanding, it seems that exposed comm time is usually over-estimated, because the first collectives will have 3~4ms more latency than expected, maybe due to the uneven workload (probably caused by flex attention's document mask) between ranks.

xmfan:
> It seems to be fixed after 20250117? (So I recently added back the compiled_autograd flag but haven't done anything yet)

Yeah the issues you mentioned should be fixed starting that rev. Currently, there is one open issue that we haven't been able to repro yet: https://github.com/pytorch/pytorch/issues/149586, but I didn't encounter it when I tried out this repo. If you're using versions before `20250415` (https://github.com/pytorch/pytorch/pull/150074), I don't think you'll get good performance from Compiled Autograd out of the box, you would need to write a compiler graph pass to get eager-like overlap. 

> Note: it's hard to move version beyond nightly 20250208 because of an unexpected performance regression: https://github.com/pytorch/pytorch/issues/147463 (even worse, there seems to be further regression after that...)

🤦 This issue seems to have gotten lost due to the labels, let me bring this up with the team.

> BTW, based on my naive understanding, it seems that exposed comm time is usually over-estimated, because the first collectives will have 3~4ms more latency than expected, maybe due to the uneven workload (probably caused by flex attention with document mask) between ranks.

Yes, if compounded, uneven workload can cause a big delta between the fastest rank and the slowest rank. For this DDP-style parallelism, ranks will compound drift for the whole forward pass and it will be shown on the first all reduce in the backward. After the first collective, the time between syncs is much lower, and we can expect less compounding. On the small nanogpt, I've seen the first gradient sync be around ~2ms and later ones be ~500us.

YouJiacheng:
I have two follow-up questions:
1. the low-contention collectives uses `torch._C._distributed_c10d._register_work(ret, Work())` and call `torch.ops._c10d_functional.wait_tensor`, so they should work well with `https://github.com/pytorch/pytorch/pull/150074` which uses `node.target != torch.ops._c10d_functional.wait_tensor.default` to skip the reorder, right?
https://github.com/pytorch/pytorch/blob/8d427e9e76eb092cfa0762fc735bc8ac0a60c1f1/test/distributed/test_symmetric_memory.py#L675-L696

2. In my understanding reordering pass will cause trouble if all waits are at the end -- DDP works in this way, and reordering pass will move the collectives right before waits and break overlapping. But here we invoke collectives without `async_op=True`, so they should be waited immediately, which won't enable overlapping even without reordering pass. Are there some other optimization passes that move these waits to the end?

YouJiacheng:
Oh I know Dynamo / torch.compile will rewrite them into `all_reduce` + `wait_tensor`, but I think `wait_tensor` will be called immediately.
But I just found a nice comment: https://github.com/pytorch/pytorch/blob/main/torch/distributed/_functional_collectives.py#L106-L132
image

YouJiacheng:
Not sure if these two passes are turned on by default:
https://github.com/pytorch/pytorch/blob/8d427e9e76eb092cfa0762fc735bc8ac0a60c1f1/torch/_inductor/comms.py#L36-L51

YouJiacheng:
Hmmm interesting, the generated code doesn't move `wait_tensor` to the end in this example:
```python
import os

import torch
from torch import Tensor
import torch._inductor.codecache
import torch._inductor.graph
from torch._logging._internal import trace_structured
import torch.distributed as dist  # type: ignore


@torch.compile
def foo(x: Tensor, y: Tensor):
    dist.all_reduce(x, dist.ReduceOp.AVG)
    z = y * y
    return x, z


rank = int(os.environ["RANK"])
world_size = int(os.environ["WORLD_SIZE"])
device = torch.device("cuda", int(os.environ["LOCAL_RANK"]))
dist.init_process_group(backend="nccl", device_id=device)


def _patched_trace_structured(name, metadata_fn, **kwargs):
    if name == "inductor_output_code":
        print(f"inductor_output_code: {metadata_fn().get('filename', 'Unknown')}")
    trace_structured(name, metadata_fn, **kwargs)
if rank == 0:
    torch._inductor.codecache.trace_structured = _patched_trace_structured  # type: ignore
    torch._inductor.graph.trace_structured = _patched_trace_structured  # type: ignore

with device:
    x = torch.ones(1024)
    y = torch.ones(1024)
    foo(x, y)

dist.destroy_process_group()

```
the generated code:
```python
# AOT ID: ['0_inference']
from ctypes import c_void_p, c_long, c_int
import torch
import math
import random
import os
import tempfile
from math import inf, nan
from cmath import nanj
from torch._inductor.hooks import run_intermediate_hooks
from torch._inductor.utils import maybe_profile
from torch._inductor.codegen.memory_planning import _align as align
from torch import device, empty_strided
from torch._inductor.async_compile import AsyncCompile
from torch._inductor.select_algorithm import extern_kernels
from torch._inductor.codegen.multi_kernel import MultiKernelCall
import triton
import triton.language as tl
from torch._inductor.runtime.triton_heuristics import start_graph, end_graph
from torch._C import _cuda_getCurrentRawStream as get_raw_stream
from torch._C import _cuda_getCurrentRawStream as get_raw_stream

aten = torch.ops.aten
inductor_ops = torch.ops.inductor
_quantized = torch.ops._quantized
assert_size_stride = torch._C._dynamo.guards.assert_size_stride
empty_strided_cpu = torch._C._dynamo.guards._empty_strided_cpu
empty_strided_cuda = torch._C._dynamo.guards._empty_strided_cuda
empty_strided_xpu = torch._C._dynamo.guards._empty_strided_xpu
reinterpret_tensor = torch._C._dynamo.guards._reinterpret_tensor
alloc_from_pool = torch.ops.inductor._alloc_from_pool
async_compile = AsyncCompile()
empty_strided_p2p = torch._C._distributed_c10d._SymmetricMemory.empty_strided_p2p


# kernel path: /tmp/torchinductor_root/sa/csaxo4lhtxz3inauixw7oe7uxur33mqj73yjegne4mnurguf6ics.py
# Topologically Sorted Source Nodes: [z], Original ATen: [aten.mul]
# Source node to ATen node mapping:
#   z => mul
# Graph fragment:
#   %mul : [num_users=1] = call_function[target=torch.ops.aten.mul.Tensor](args = (%arg1_1, %arg1_1), kwargs = {})
triton_poi_fused_mul_0 = async_compile.triton('triton_poi_fused_mul_0', '''
import triton
import triton.language as tl

from torch._inductor.runtime import triton_helpers, triton_heuristics
from torch._inductor.runtime.triton_helpers import libdevice, math as tl_math
from torch._inductor.runtime.hints import AutotuneHint, ReductionHint, TileHint, DeviceProperties
triton_helpers.set_driver_to_gpu()

@triton_heuristics.pointwise(
    size_hints={'x': 1024}, 
    filename=__file__,
    triton_meta={'signature': {'in_ptr0': '*fp32', 'out_ptr0': '*fp32', 'xnumel': 'i32', 'XBLOCK': 'constexpr'}, 'device': DeviceProperties(type='cuda', index=0, multi_processor_count=132, cc=90, major=9, regs_per_multiprocessor=65536, max_threads_per_multi_processor=2048, warp_size=32), 'constants': {}, 'configs': [{(0,): [['tt.divisibility', 16]], (1,): [['tt.divisibility', 16]], (2,): [['tt.divisibility', 16]]}]},
    inductor_meta={'grid_type': 'Grid1D', 'autotune_hints': set(), 'kernel_name': 'triton_poi_fused_mul_0', 'mutated_arg_names': [], 'optimize_mem': True, 'no_x_dim': False, 'num_load': 1, 'num_reduction': 0, 'backend_hash': '74EE655CF3D3704CBC7ACD828702DCAEA7302FC9D60FB5498F16887EB8564E9A', 'are_deterministic_algorithms_enabled': False, 'assert_indirect_indexing': True, 'autotune_local_cache': True, 'autotune_pointwise': True, 'autotune_remote_cache': None, 'force_disable_caches': False, 'dynamic_scale_rblock': True, 'max_autotune': False, 'max_autotune_pointwise': False, 'min_split_scan_rblock': 256, 'spill_threshold': 16, 'store_cubin': False},
    min_elem_per_thread=0
)
@triton.jit
def triton_poi_fused_mul_0(in_ptr0, out_ptr0, xnumel, XBLOCK : tl.constexpr):
    xnumel = 1024
    xoffset = tl.program_id(0) * XBLOCK
    xindex = xoffset + tl.arange(0, XBLOCK)[:]
    xmask = xindex < xnumel
    x0 = xindex
    tmp0 = tl.load(in_ptr0 + (x0), xmask)
    tmp1 = tmp0 * tmp0
    tl.store(out_ptr0 + (x0), tmp1, xmask)
''', device_str='cuda')


async_compile.wait(globals())
del async_compile

def call(args):
    arg0_1, arg1_1 = args
    args.clear()
    assert_size_stride(arg0_1, (1024, ), (1, ))
    assert_size_stride(arg1_1, (1024, ), (1, ))
    with torch.cuda._DeviceGuard(0):
        torch.cuda.set_device(0)
        # Topologically Sorted Source Nodes: [tensor], Original ATen: [_c10d_functional.all_reduce]
        torch.ops._c10d_functional.all_reduce_.default(arg0_1, 'avg', '0')
        # Topologically Sorted Source Nodes: [wait_tensor], Original ATen: [_c10d_functional.wait_tensor]
        torch.ops._c10d_functional.wait_tensor.default(arg0_1)
        del arg0_1
        buf4 = empty_strided_cuda((1024, ), (1, ), torch.float32)
        # Topologically Sorted Source Nodes: [z], Original ATen: [aten.mul]
        stream0 = get_raw_stream(0)
        triton_poi_fused_mul_0.run(arg1_1, buf4, 1024, stream=stream0)
        del arg1_1
    return (buf4, )


def benchmark_compiled_module(times=10, repeat=10):
    from torch._dynamo.testing import rand_strided
    from torch._inductor.utils import print_performance
    arg0_1 = rand_strided((1024, ), (1, ), device='cuda:0', dtype=torch.float32)
    arg1_1 = rand_strided((1024, ), (1, ), device='cuda:0', dtype=torch.float32)
    fn = lambda: call([arg0_1, arg1_1])
    return print_performance(fn, times=times, repeat=repeat)


if __name__ == "__main__":
    from torch._inductor.wrapper_benchmark import compiled_module_main
    compiled_module_main('None', benchmark_compiled_module)

```

YouJiacheng:
Very interesting.
Case 1: return the average
`wait` is NOT pushed to the end
```python
@torch.compile
def foo(x: Tensor, y: Tensor):
    x_avg = fcol.all_reduce(x, "avg", "0")
    y_sq = y * y
    return x_avg, y_sq

```
```python
def call(args):
    arg0_1, arg1_1 = args
    args.clear()
    assert_size_stride(arg0_1, (1024, ), (1, ))
    assert_size_stride(arg1_1, (1024, ), (1, ))
    with torch.cuda._DeviceGuard(0):
        torch.cuda.set_device(0)
        buf0 = empty_strided_cuda((1024, ), (1, ), torch.float32)
        # Topologically Sorted Source Nodes: [tensor], Original ATen: [_c10d_functional.all_reduce]
        stream0 = get_raw_stream(0)
        triton_poi_fused_all_reduce_0.run(arg0_1, buf0, 1024, stream=stream0)
        del arg0_1
        # Topologically Sorted Source Nodes: [tensor], Original ATen: [_c10d_functional.all_reduce]
        torch.ops._c10d_functional.all_reduce_.default(buf0, 'avg', '0')
        # Topologically Sorted Source Nodes: [out_0], Original ATen: [_c10d_functional.wait_tensor]
        torch.ops._c10d_functional.wait_tensor.default(buf0)
        buf5 = empty_strided_cuda((1024, ), (1, ), torch.float32)
        # Topologically Sorted Source Nodes: [out_1], Original ATen: [aten.mul]
        stream0 = get_raw_stream(0)
        triton_poi_fused_mul_1.run(arg1_1, buf5, 1024, stream=stream0)
        del arg1_1
    return (buf0, buf5, )
```
Case 2: assign the average to an attribute
`wait` IS pushed to the end
```python
@torch.compile
def foo(x: Tensor, y: Tensor):
    x.avg = fcol.all_reduce(x, "avg", "0")
    y_sq = y * y
    return None, y_sq
```
```python
def call(args):
    arg0_1, arg1_1 = args
    args.clear()
    assert_size_stride(arg0_1, (1024, ), (1, ))
    assert_size_stride(arg1_1, (1024, ), (1, ))
    with torch.cuda._DeviceGuard(0):
        torch.cuda.set_device(0)
        buf0 = empty_strided_cuda((1024, ), (1, ), torch.float32)
        # Topologically Sorted Source Nodes: [tensor], Original ATen: [_c10d_functional.all_reduce]
        stream0 = get_raw_stream(0)
        triton_poi_fused_all_reduce_0.run(arg0_1, buf0, 1024, stream=stream0)
        del arg0_1
        # Topologically Sorted Source Nodes: [tensor], Original ATen: [_c10d_functional.all_reduce]
        torch.ops._c10d_functional.all_reduce_.default(buf0, 'avg', '0')
        buf3 = empty_strided_cuda((1024, ), (1, ), torch.float32)
        # Topologically Sorted Source Nodes: [z], Original ATen: [aten.mul]
        stream0 = get_raw_stream(0)
        triton_poi_fused_mul_1.run(arg1_1, buf3, 1024, stream=stream0)
        del arg1_1
        # Topologically Sorted Source Nodes: [wait_tensor], Original ATen: [_c10d_functional.wait_tensor]
        torch.ops._c10d_functional.wait_tensor.default(buf0)
    return (buf3, buf0, )
```
Case 3: Case 2 but `torch._inductor.config.reorder_for_locality = False`
`wait` is NOT pushed to the end
```python
def call(args):
    arg0_1, arg1_1 = args
    args.clear()
    assert_size_stride(arg0_1, (1024, ), (1, ))
    assert_size_stride(arg1_1, (1024, ), (1, ))
    with torch.cuda._DeviceGuard(0):
        torch.cuda.set_device(0)
        buf0 = empty_strided_cuda((1024, ), (1, ), torch.float32)
        # Topologically Sorted Source Nodes: [tensor], Original ATen: [_c10d_functional.all_reduce]
        stream0 = get_raw_stream(0)
        triton_poi_fused_all_reduce_0.run(arg0_1, buf0, 1024, stream=stream0)
        del arg0_1
        # Topologically Sorted Source Nodes: [tensor], Original ATen: [_c10d_functional.all_reduce]
        torch.ops._c10d_functional.all_reduce_.default(buf0, 'avg', '0')
        # Topologically Sorted Source Nodes: [wait_tensor], Original ATen: [_c10d_functional.wait_tensor]
        torch.ops._c10d_functional.wait_tensor.default(buf0)
        buf5 = empty_strided_cuda((1024, ), (1, ), torch.float32)
        # Topologically Sorted Source Nodes: [y_sq], Original ATen: [aten.mul]
        stream0 = get_raw_stream(0)
        triton_poi_fused_mul_1.run(arg1_1, buf5, 1024, stream=stream0)
        del arg1_1
    return (buf5, buf0, )
```

Full code:
```python
import os

import torch
from torch import Tensor
import torch._inductor.codecache
import torch._inductor.graph
from torch._logging._internal import trace_structured
import torch.distributed as dist
import torch.distributed._functional_collectives as fcol


# torch._inductor.config.reorder_for_locality = False
@torch.compile
def foo(x: Tensor, y: Tensor):
    x_avg = fcol.all_reduce(x, "avg", "0")
    y_sq = y * y
    return x_avg, y_sq


# @torch.compile
# def foo(x: Tensor, y: Tensor):
#     x.avg = fcol.all_reduce(x, "avg", "0")
#     y_sq = y * y
#     return None, y_sq


rank = int(os.environ["RANK"])
world_size = int(os.environ["WORLD_SIZE"])
device = torch.device("cuda", int(os.environ["LOCAL_RANK"]))
dist.init_process_group(backend="nccl", device_id=device)


def _patched_trace_structured(name, *args, **kwargs):
    if args:
        metadata_fn, *_ = args
    else:
        metadata_fn = kwargs.get("metadata_fn", lambda: {})
    if name == "inductor_output_code":
        print(f"inductor_output_code: {metadata_fn().get('filename', 'Unknown')}")
    trace_structured(name, *args, **kwargs)
if rank == 0:
    torch._inductor.codecache.trace_structured = _patched_trace_structured  # type: ignore
    torch._inductor.graph.trace_structured = _patched_trace_structured  # type: ignore

with device:
    x = torch.ones(1024)
    y = torch.ones(1024)
    foo(x, y)

dist.destroy_process_group()

```

YouJiacheng:
Okay I know why: 
If `visit` is called in sequence `[(output, wait_tensor), (output, mul)]` then it will insert `wait_tensor` right before `output` then insert `mul` right before `output` and break overlapping.
In another word, if `output` has multiple producers, then the order of these producers can vary with subtle changes.
In backward, this is safe, since all producers of `output` will be `wait_tensors`.

xmfan:
I'm gonna close this PR now that we've looked at it. There's an issue where `torch._dynamo.config.compiled_autograd = True` won't work: https://github.com/pytorch/pytorch/issues/152219. You can use this PR as a reference for the alternative.

PR #108, opened 2025-07-13
Align training batches to bos token (New WR 173.4s), incl prior changes from @vagrawal

ClassicLarry:
1. Included engineering optimizations from https://github.com/KellerJordan/modded-nanogpt/pull/102 by @vagrawal
2. Updated distributed_data_generator() to align to bos_token for training data. Specifically: Added a helper function in the distributed data loader that returns the next {world_size} starting points, one for each gpu, such that points are spaced at least {local_batch_size} distance from each other, all start with a bos_token, keep same order, and skip the minimum number of tokens. Justification is right now a nontrivial number of samples start mid-sample, learning from mid-sample context is more challenging, and not having a bos_token in the intra-document context window may degrade attention sink behavior. Validation data kept consistent with prior records. Since the validation batch_size is 5.3x larger than training batch_size, validation data has 5.3x fewer samples that start mid-sample.
3. As a follow-on from 2: Reduced number of iterations from 1770 to 1750, increased cooldown frac from 0.4 to 0.45, and decreased minimum lr schedule factor from 0.1 to 0.05.

Discussion where prior record holders consider bos_token change within spirit of the challenge: https://github.com/KellerJordan/modded-nanogpt/discussions/23#discussioncomment-13689953

Validated results with 20 runs on 8H100 GPUs with p<0.01:
```
import scipy.stats
import torch

accs = [
    3.2784,3.2791,3.2819,3.2801,3.2788,3.2794,3.2782,3.277,3.2784,3.2773,
    3.2803,3.2792,3.2792,3.2804,3.2817,3.2805,3.2783,3.2789,3.2779,3.2778]
times = [
    173.368,173.424,173.291,173.441,173.397,173.348,173.402,173.463,173.209,173.338,
    173.291,173.377,173.322,173.287,173.265,173.375,173.331,173.413,173.384,173.462]

# p=0.0049

print('acc:',torch.std_mean(torch.tensor(accs)))
# acc: (tensor(0.0013), tensor(3.2791))

print('time:',torch.std_mean(torch.tensor(times)))
# time: (tensor(0.0682), tensor(173.3594))
```


vagrawal:
Are you able to get `loss < 3.28` with statistical significance? I got `loss > 3.28` with `p < 0.05` in the `main` branch. See [my comment ](https://github.com/KellerJordan/modded-nanogpt/pull/102#issuecomment-2931962347). You could potentially save ~10 more iterations to reach the loss of `main` branch.

ClassicLarry:
> Are you able to get `loss < 3.28` with statistical significance? I got `loss > 3.28` with `p < 0.05` in the `main` branch. See [my comment ](https://github.com/KellerJordan/modded-nanogpt/pull/102#issuecomment-2931962347). You could potentially save ~10 more iterations to reach the loss of `main` branch.

Yes, for this pull req loss is under 3.28 with p=0.0049 based on 20 runs. I only ran main branch once, interesting to learn its avg loss may be higher. Imo its preferable to keep the record meaningfully below 3.28 so people can test ML changes without requiring 80+ runs to validate p values, to keep the challenge financially accessible.

KellerJordan:
Great stuff I am starting to evaluate.

KellerJordan:
Accepting as new record. I also retimed with torch==2.9.0.dev20250713+cu126, where it got 171915ms, which will be the record time

KellerJordan:
btw, do you happen to know how much each of the two changes improved things - aligning with EoS, and increasing cooldown from .4 to .45?

ClassicLarry:
> btw, do you happen to know how much each of the two changes improved things - aligning with EoS, and increasing cooldown from .4 to .45?

I dont on this. My hunch is that .45 cooldown is probably detrimental on its own. From the eos change it was pretty clear that some small retuning could drop at least 20 iterations. Dropping the learning rate earlier seemed to give near term gains at the expense of long term slow downs. Since the eos change brings the general loss lower, those near term gains are enough to get over the 3.28 threshold. This hunch could be falsified with a simple test though.

PR #114, opened 2025-07-27
Rename records folders to YYYY-MM-DD format for chronological sorting

kiankyars:
QOL PR to make it easier to go through previous records.


kiankyars:
@ClassicLarry please look at this

ClassicLarry:
I agree this is a good direction, given that the number of features in this repo makes readability a pain point, and the Oct2025 records are now mixing with the Oct2024 record folder. If you could update this for all folders, add separate subfolders under records/ for track_1_short, track_2_medium, and update the readme log links in the two tables, I'll merge. In my opinion the main value of this repo is as a learning tool so this change will be great!

kiankyars:
> I agree this is a good direction, given that the number of features in this repo makes readability a pain point, and the Oct2025 records are now mixing with the Oct2024 record folder. If you could update this for all folders, add separate subfolders under records/ for track_1_short, track_2_medium, and update the readme log links in the two tables, I'll merge. In my opinion the main value of this repo is as a learning tool so this change will be great!

Yes, will do!

kiankyars:
I made the changes you requested and I would also recommend to change the file name of all of the medium records by removing medium from the log since it's now redundant because it's included in the subdirectory, but let me know what you think.

ClassicLarry:
Good idea!

kiankyars:
Made the change

ClassicLarry:
I am seeing https://github.com/KellerJordan/modded-nanogpt/tree/d59fcbbebe2eb4866e9123cf35fa646ff85df445/records/track_1_short/2025-06-15_GPT2MediumOptimizationLeaderboard is under short track, can move to medium. Otherwise looks good.

kiankyars:
Thanks, I just saw that, Didn't catch that because I guess it's not in the readme yet

ClassicLarry:
thanks, people will appreciate your update.

Gusarich:
@ClassicLarry I think it'll be better for future PRs to squash commits on merge so that the main branch history doesn't get flooded

PR #118, opened 2025-08-27
New WR: Flash Attention 3 and Batch Dimension (166.1 seconds). Includes WR Changes from PR#109 and PR#117

varunneal:
# New record 08/27/25

This submission includes recent WR changes by @ClassicLarry [(08/23/25)](https://github.com/ClassicLarry/modded-nanogpt/tree/master/records/082325_SparseAttnGate) and @byronxu99 [(07/18/25)](https://github.com/KellerJordan/modded-nanogpt/pull/109). 

The main idea of this record is to use input tensors with `batch_size > 1` throughout our training run. Increasing  `batch_size` increases GPU utilization and allows us to use shorter input sequences for training.  However, since Flex Attention's is inefficient for `batch_size > 1`, we use [Flash Attention v3](https://github.com/Dao-AILab/flash-attention). 
The official version of this module is incompatible with `torch.compile` and causes graph breaks.  However, a [recent PR](https://github.com/Dao-AILab/flash-attention/pull/1769) by  [@guilhermeleobas](https://github.com/guilhermeleobas) addresses this issue.

## Timing and Validation

Validated over 7 runs:
- In 1695 training steps, this run achieves a loss <3.28 (`p=0.0031`) 
- In 166.10 seconds on average, or <166.25 seconds (`p=0.0024`), 

```
import scipy.stats
import torch
import numpy as np

accs = [
    3.2769, 3.2782, 3.2790, 3.2791, 3.2791, 3.2780, 3.2782
]

times = [
    166.247, 166.117, 165.977, 166.135, 166.045, 166.044, 166.157
]

# p=0.0008

print('p=%.4f' % scipy.stats.ttest_1samp(times, 166.25, alternative='less').pvalue)
# p=0.0024

print(f"{np.mean(times):.4f}")
# 166.1031
```

In my timing, this is a 2.1 second mean improvement over https://github.com/KellerJordan/modded-nanogpt/pull/117. 
Given the comfortable p-value, the number of steps can also probably be brought down by 5-15 while achieving loss <3.28.

I used SXM5 8 x H100 via Prime Intellect for validation compute. 

### Motivation

PyTorch's Flex Attention experiences a slowdown >10% wallclock for inputs with `batch_size > 1`. As such, previous records would train on very long sequence lengths (`48 * 1024`) with no batch dimension. Attention is approximately `O(|seq_len|^2 x |batch_size|)`, so this is theoretically bad, but it was mitigated by using aggressive blocking masking.
Attention used a `block_mask` which only grew at most to `1664` tokens (and was often shorter due to document masking). However, GPU utilization for attention is higher when tokens are distributed along the batch dimension.

Additionally, increasing the batch size allows us to decrease sequence length while maintaining the total number of tokens processed per step. WR#26 by @ClassicLarry found that validation loss decreases when we train only on sequences beginning with the Beginning of Sequence token (`<BoS>`). Decreasing the sequence length ensures makes it more likely that `<BOS>` is present in the attention window. In order generate batches where each sequence begins with `<BOS>`, I have created the helper class `EOSBatchFinder`. This class pre-indexes shards with the location of `<BoS>` for slight speedups. 

### Flash Attention 3

Most of the Hopper-specific benefits in Flash Attention 3 are incorporated into PyTorch's Flex Attention already. However, the latter implementation is fastest with `batch_size == 1`, Flash Attention 3 is as fast as Flex Attention for 1 dimensional input sequences, and increases in speed as we distribute tokens along the batch dimension.  I measured a 9% wallclock decrease for FA3 when using an optimal ratio of batch dimension to sequence length (`24: 2048`) over a single batch dimension (`1: 49152`) (on a single Hopper H100). 

As mentioned above, we need to use an unmerged PR in order to use FA3 with `torch.compile`. 
You can build the wheel like so:

```
pip install -U pip wheel setuptools ninja numpy packaging psutil

git clone https://github.com/guilhermeleobas/flash-attention.git
cd flash-attention/hopper
git switch guilhermeleobas/fa3-compile

export MAX_JOBS=32                             # Can increase based on machine
export FLASH_ATTENTION_FORCE_BUILD=TRUE        # skip prebuilt wheel fetch
export FLASH_ATTENTION_DISABLE_SM80=TRUE       # Hopper-only
export FLASH_ATTENTION_DISABLE_FP16=TRUE       # leave BF16, FP8
export FLASH_ATTENTION_DISABLE_HDIM64=TRUE     # NanoGPT only uses HDIM = 128
export FLASH_ATTENTION_DISABLE_HDIM96=TRUE
export FLASH_ATTENTION_DISABLE_HDIM192=TRUE
export FLASH_ATTENTION_DISABLE_HDIM256=TRUE

python setup.py bdist_wheel
```

Additionally, I have uploaded a prebuilt wheel  [here](https://github.com/varunneal/flash-attention/releases/tag/v3.0.0b1-alpha), though it will likely be faster to build it yourself than download this wheel.

For exact reproduction, I recommend that you install Torch Nightly 2.9.0.dev20250718 and install the FA3 wheel afterward:

```
pip install --pre "torch==2.9.0.dev20250718+cu126" --index-url https://download.pytorch.org/whl/nightly/cu126

# typical path to FA3 Wheel
pip install flash-attention/hopper/dist/flash_attn_3-3.0.0b1-cp39-abi3-linux_x86_64.whl
```

For me, Torch Nightly 2.9.0.dev20250713 was incompatible with PR#109.

### Attention Masks

Unfortunately, Flash Attention does not support complex Block Masks like Flex Attention. Therefore, `create_blockmasks` was removed. Instead, we only are given the parameter `window_size` where we can specify the number of left tokens to attend to.

I kept the existing long-short sliding window block mask pattern, as well as the idea that the window sizes should linearly increase over the length of the training run. To aid with this, I modified `get_lr(step)` to instead be `get_lr_and_ws(step)`. Additionally, I added a hyperparameter `ws_schedule` which specifies what the longer window size should be during each portion of the run. I additionally added the size of blocks in a window as a hyperparameter `bandwidth=128`. 

I have picked a linear schedule with three steps: `ws_schedule=(3, 7, 11)`.  Currently, `torch.compile` creates a new compilation graph per each step in `ws_schedule`. Therefore, each graph needs to be warmed up separately. I have increased the number of warmup steps from `10` to `60`. The compile time is dominated by the first iteration so this will take approximately `len(ws_schedule)` times longer than before.

Removing document masking had a noticeably negative impact on validation loss, however the benefits of a short sequence length counteract this.

### Potential Improvements

- Batch size scheduling: Previously, the block mask acted as a proxy for batch size. Now block size can be controlled explicitly and sequenced according to critical batch size theory. I have added code in `distributed_data_generator` that allows for changing the batch size and sequence length yielded after the generator is created. 
- The current block mask window schedule `(3, 7, 11)` can almost certainly  be improved upon.
- Hyperparameter tuning might change with smaller sequence length. Rotary base, validation sequence length, learning rates etc. should be re-tuned. I haven't done that for this run. 
- FA3 has additional features over Flex Attention that may be useful.


ClassicLarry:
Wow this is neat! Great implementation of a bleeding edge tech. A lot to digest here. This improvement opens a lot of doors for what's possible.

First on the loss side:
Your loss on step 250 is coming in way lower at 3.9787, compared to 4.07 before. I wonder how much of this is due to having a constant attention size of 384 for this whole duration, instead of 128->256. I am surprised by how much the shorter and aligned sequence lengths appears to help.

On runtime, measured from the cpu (spikes every 125 are from cpu timer around validation loss check):
Standard Flex attention for reference with staggered linear increase in attention window. 
image

Zooming in on Flash Attention step times. Biggest differentiation is on the right third, as expected.
image

Zooming out, the flash attention implementation is incurring a .3-.4s penalty every 172 steps or so.

image
 I assume this is when EOSBatchFinder() loads in the next shard, and why there is a .5s penalty on step1. I may be misinterpretting some of the cpu vs gpu timing aspects here. Without synchronizing the gpu its hard to tell. Now that you have batched this computation, I wonder if there is a way to distribute it reasonably across the GPUs, or space this computation into the portions where the GPUs are least active.

YouJiacheng:
IIUC, you can wrap it with a custom op and make `window_size` to be an internal state of it to avoid re-compilation.
Actually you can just wrap the official version of FA3 with a custom op.
You may also try wrapping the `window_size` as a CPU tensor. it might work.
```python
global_variable_window_size = 3
@torch.library.custom_op("xxx::yyy", mutates_args=(), device_types="cuda")
def stateful_flash_attn(q, k, v):
    flash_attn_func(..., window_size=(global_variable_window_size, 0))

# register fake, backward etc.
```

YouJiacheng:
> Removing document masking had a noticeably negative impact on validation loss, however the benefits of a short sequence length counteract this.

I don't understand, do you mean a shorter sequence length improves validation loss?

YouJiacheng:
I think you can just use the varlen version to get back the document mask.
image

ClassicLarry:
> > Removing document masking had a noticeably negative impact on validation loss, however the benefits of a short sequence length counteract this.
> 
> I don't understand, do you mean a shorter sequence length improves validation loss?

My hypothesis is that the primary reason for the loss improvement, which is very substantial on step 250, is a 25% higher sequence density per step, which provides better coverage over the token space, better coverage over the conceptual space of fineweb, and more accurate gradients, for faster and more stable training. Based on the plot above it looks like 10 shards are being processed instead of 8, because sequences are getting truncated at a max distance of 2048 instead of 49152. 

Fineweb shard 1, roughly 100M tokens and 144719 sequences, has a mean sequence length of 690, max sequence length of 78977, and min of 42.

image

Here's what the 70,000+ token sequences look like:
Example 1:
```
num tokens: 76142. num_chars: 350229
[:100]  <|endoftext|>UNITED STATES of America, Plaintiff–Appellee, v. Mohammad EL–MEZAIN; Ghassan Elashi; Sh
[10000:10100]  d to Hamas even though HLF continued to support the same zakat committees that Hamas controlled.
Fro
[20000:20100]  rorist organization any of the zakat committees or anyone connected to the committees. They argued t
[30000:30100]  tive light on both the witness and the military operation. The defense also elicited from Major Lior
```

Example 2:
```
num tokens: 74860. num_chars: 247780
[:100]  <|endoftext|>~The Fox Den~
He clutched his shoulder. Slow motion fingers found nothing. No muscle, n
[10000:10100]   movie are we watching?”
“Uh… d’you mind if we watch volleyball. Only I found a channel showing a to
[20000:20100]  for the fans and then autographs and selfies. It was never a productive session, but everyone recogn
[30000:30100]  dy lived here.” She fluttered her eyelashes at Suna. “You must have been exceptionally good to have 
```

Example 3:
```
num tokens: 78977. num_chars: 396302
[:100]  <|endoftext|>(22 May 2013)
It is our aim to inform citizens about the principles and rules governing
[10000:10100]  a city. The city of Antwerp and the province received investment worth some EUR 80 million in the cu
[20000:20100]  an economy. The Commission has recently created an extensive database of best urban development good
[30000:30100]  egy in our next generation of Regional funds programming and firmly embedded in the national, region
```

From the examples above, it looks like the longer sequences have narrow conceptual scope and skewed token probabilities. If a step gets caught updating on a single Plaintiff case for its entire step, its going to have a very poor estimate of the true gradient on the validation set. Since the attention window is only ~1600 tokens anyways, we get a better estimate of the global token and concept distribution if we sample two sequences of size 1600, rather than 1 sequence of size 3200. My hunch is that when we scale down the max seq length from 49152 to 2048 this effect becomes far more pronounced compared to having the bos_token in the attention window.

varunneal:
Thank you both for looking over my PR!

> I  don't understand, do you mean a shorter sequence length improves validation loss?

ClassicalLarry gave a great answer above. What I meant was simply that shorter aligned sequence lengths are very likely to include the BOS token at the very start. I believe the point he raised above is that sampling at a lower sequence length allows us to sample a great diversity of sequences in a single batch step. This is not a point I had considered but might be the strongest effect. 

> I think you can just use the varlen version to get back the document mask.

`flash_attn_varlen_func` takes in input `q, k, v` of shape `(# tokens, nheads, headdim)`, e.g. it doesn't support a batch dimension

> IIUC, you can wrap it with a custom op and make window_size to be an internal state of it to avoid re-compilation.
> Actually you can just wrap the official version of FA3 with a custom op.
> You may also try wrapping the window_size as a CPU tensor. it might work.

I will take a look at these various approaches! Hopefully there is something more elegant than multiple compilation graphs.

YouJiacheng:
> `flash_attn_varlen_func` takes in input `q, k, v` of shape `(# tokens, nheads, headdim)`, e.g. it doesn't support a batch dimension

it supports a batch dim.
it packs all sequences in a batch into a buffer, and `cu_seqlens_*` has the shape of `(batch_size + 1,)`.
a 1-dimensional buffer plus `cu_seqlens_*` implement a container for a batch of sequences with different lengths.
you can check the implementation of varlen.

YouJiacheng:
I show the CuTE DSL implementation cuz it's easier to read.
You still use a batch, but because the sequences have different lengths, you must use a jagged tensor, which can be implemented by `values` (or, a buffer) plus `offsets` (or, `cu_seqlens_*`)
image
image
image
https://docs.pytorch.org/FBGEMM/fbgemm_gpu/overview/jagged-tensor-ops/JaggedTensorOps.html

YouJiacheng:
> What I meant was simply that shorter aligned sequence lengths are very likely to include the BOS token at the very start.

But with document mask, we effectively ensure BOS token to be the first token for all documents (sequences) except the first one.

ClassicLarry:
My main modeling update from this PR is that correlated data samples is a big problem for language models when the tokens per step is not substantially greater than the document length. I'd be curious if frontier labs have a sophisticated data preprocessing layer, where samples get truncated and shuffled around, such that the data for each step closely mirrors the target distribution on a token and concept level.

ClassicLarry:
> > What I meant was simply that shorter aligned sequence lengths are very likely to include the BOS token at the very start.
> 
> But with document mask, we effectively ensure BOS token to be the first token for all documents (sequences) except the first one.

The BOS token was getting dropped from the context window for longer sequences since the attention window is only 128-1600. Document mask wasn't solving this, but truncating the sequence length does.

YouJiacheng:
> The BOS token was getting dropped from the context window for longer sequences since the attention window is only 128-1600. Document mask wasn't solving this, but truncating the sequence length does.

Yep but this is irrelevant to "include the BOS token at the very start"

varunneal:
> Yep but this is irrelevant to "include the BOS token at the very start"

Sorry, I mean to say that with short sequences, the likelihood that a given token gets to attend to the BOS token at the very start is much higher. Based on ClassicLarry's findings, I believe this will enhance the ability of the BOS token to act as an attention sink.

> `cu_seqlens_*` has the shape of (batch_size + 1,)

From what I understand, "batch size" in the varlen API means "number of documents". Each entry of `cu_seqlens_*` is a token offset into the flattened Q/K/V. As far as I know, FA does not support `cu_seqlen_`  being a Jagged Tensor. It is expected to be a packed 1D tensor of concatenated sequences. 

> Now that you have batched this computation, I wonder if there is a way to distribute it reasonably across the GPUs, or space this computation into the portions where the GPUs are least active.

This is an interesting idea. I analyzed the logs and found:
- We have a spike at steps 1, 173, 345, 519, 691, 863, 1036, 1208, 1381, and 1552
- On average this adds 284.5ms 10 times -> 2.85 seconds in total
- Loading the shard from disk dominates the spike time. On an H100 this load takes 220ms.

YouJiacheng:
> From what I understand, "batch size" in the varlen API means "number of documents". Each entry of `cu_seqlens_*` is a token offset into the flattened Q/K/V.

Yes, this understanding is correct. So this is exactly for a batch of sequences. So wdym by "it doesn't support a batch dimension"?
GPU computation will be performed independently across different documents, just like a regular batch.
If you want to avoid excessively long documents, you can just split them.
I can't see why you think this will be less efficient than an equal-length batch.

For jagged tensor I mean, `q + cu_seqlens_q` is equivalent to a jagged tensor.

YouJiacheng:
>  It is expected to be a packed 1D tensor of concatenated sequences.

`cu_seqlens_*` are just simple 1D tensors of the shape `(batch_size + 1,)`, nothing is concatenated.
`q, k, v` are packed tensors of concatenated sequences.

varunneal:
Ah got it, you are proposing using Flash Attention's `flash_attn_varlen_func` to recreate the functionality of the previous 1D-tensor + document masking logic. This is a different direction than my PR, which is proposing fixed sized batches. I cannot speak to the GPU utilization for fixed vs variable sized batches (I have not tested it) but it would be an interesting area of exploration. I believe that the idea of truncating long documents could be easily implemented by using this PR's data generator for small cutoff sequence length and reshaping it into 1 dimension. 

In the case of this record with fixed batch sizes, I think document masking would be useful if it could be done per sequence. E.g. for an input tensor (B, T), we could mask each row according to document boundaries. That is currently not possible in Flash Attention.

varunneal:
> If you want to avoid excessively long documents, you can just split them.

Taking a look at this approach @YouJiacheng. Since some documents will be shorter than the max seq length, we will get a variable number of tokens per step, though this might be worth it. Also fixing a bug with EOSBatchFinder (previously I had placed token at the end of sequences, changing to BOS to be correct and in-line with ClassicalLarry's research)

ClassicLarry:
Some ablations that would be informative here:
1. Prior record with sequence jump to Bos token after every 2048.
2. 1 plus update attn sliding window schedule to new 3 interval setup
3. 2 plus remove document masking
4. Any other ML updates im missing

If the remaining changes are all engineering optimizations, this should help catch any implementation bugs if loss from ablation 3 equals loss from this record. I’ll prob end up running these tests at some point out of curiosity.

varunneal:
I will compile each of these implementations and record the losses. One candidate for (4) is that FA3 might have slightly different internals than Flex Attention, including the non-deterministic backward pass.

varunneal:
@YouJiacheng do you have advice on how to deal with varlen sequences and the requirement that the FP8 mm_op desires tensor shape to be divisible by 16? Also that `torch.compile` generally is not happy with dynamically shaped inputs? I'm considering padding.

YouJiacheng:
my proposal originally is about variable batch size (and fixed number of tokens per step), so you only need to pad `cu_seqlens_*` to get static shape.

variable seqlens with fixed batch size, will require dynamic shape or otherwise we need excessively many compilation…

dynamic shape will degrade performance.

varunneal:
@ClassicLarry I got the following result:

| run         | val@250 | val@875 | val@1695 |
| ------------- | ------- | ------- | -------- |
| PR 117 [^1]  | 4.0701  | 3.5341  | 3.2770   |
| PR 117 [^2] | 4.0788  | 3.5355  | 3.2772   |
| ablation 1    | 4.0532  | 3.5262  | 3.2748   |
| ablation 2    | 3.9351  | 3.5224  | 3.2761   |
| ablation 3    | 3.9991  | 3.5299  | 3.2815   |
| PR 118  [^3]      | 3.9738  | 3.5270  | 3.2791   |

[^1]: Using your`020630eb-2191-4ba2-9ee4-4cdc94316943.txt` as reference for PR 117
[^2]: Rerun on a 1 x H100 by myslf
[^3]: Using my `1d46fee6-b32c-48de-bd61-0a326442ec4e.txt` as reference

You can audit what I've done [here](https://github.com/varunneal/modded-nanogpt/tree/ablations/ablation-logs). The numbers are close together and I only ran each experiment once, so it is hard to have full confidence. I tentatively want to say:
- Capping the max document length helps a bit 
- The sliding window schedule ((3, 1), (7, 3), (11, 5)) is very good at the start but evens out to roughly the same performance (imo indicates we can tune the schedule for even greater performance)
- Removing document masking hurts a bit

varunneal:
Addendum to above:

I discovered yesterday that my PR's Data Generator includes **no** token 50256. This was an accident. However, since my PR appears to have lower loss than Ablation 3 in the above, I re-ran Ablation 3 but dropped the BOS token: 

run | val@250 | val@875 | val@1695
-- | -- | -- | --
ablation 3 | 3.9991 | 3.5299 | 3.2815
ablation 3 w/ no BOS | 4.0013 | 3.5317 | 3.2830
PR 118  | 3.9738 | 3.5270 | 3.2791

The sample size of 1 is just not enough to know what's going on, though I will update my PR to include the BOS token (& I will probably clean up the data generator -- I wrote it quite poorly) and I will try to implement some changes suggested by @YouJiacheng). After that I will re-measure on the 8 x H100 with more samples

ClassicLarry:
Thanks! Takeaways are
1. slight but very real improvement from truncating document length, perhaps due to increased diversity in data per step.
2. slight setback from removing document mask, less than I expected.
3. negligible setback in loss from attention window schedule update, but also slight/negligible speedup in runtime- analysis below.
4. some potentially unexplained gap between ablation 3 w/ no BOS and PR 118, beyond the effect of the stochastic backwards pass.

Plotting the attention window schedule:
image
The validation score follows the sliding window schedule, so there isn't an apples-to-apples comparison on mid-training model capability. Comparing old vs new:

- step 125: [128 window, 3.6 loss->384 window, 3.3 loss]
- step 250: [256 window, 4.07 loss->384 window, 3.93 loss]
- step 875: [896 window, 3.53 loss->896 window, 3.52 loss]
- step 1695: [1792 window, 3.28 loss->1408 window, 3.28 loss]

The improved performance on step 250 may be largely if not completely an illusion, due to the longer context window (~3 sentence context instead of 2) for validation. By step 875 when the windows match, the losses are almost equivalent, and the improved loss of the new schedule is likely cancelled out by the increase in runtime from having a slightly longer average context in the first 875 steps. The final model from ablation 2 is given less context than ablation 1. It might perform equivalent or better with the same validation context length. Overall I don't see evidence that the new attention window schedule is better, but my understanding is it was updated for compiler reasons.

As shown [above](https://github.com/KellerJordan/modded-nanogpt/pull/118#issuecomment-3227154552), training on longer attention windows is slower. Some of the training speedup is coming from a 3% lower attention window size on average, that finishes at 1408 instead of 1792. It would be interesting to see how, for a given trained model, validation loss varies as a function of the validation attention window. This is super easy to tune since its post training. Alternatively, there could be some gains from scaling the final window size up, maybe to 2048, right near the end of training.

One other effect- The train and test distributions dont match, and haven't for many records, ever since the train seq length (2048 or 49152) diverged from validation seq length (262144). The training and validation context windows both vary from 0 to 1408, but the training distribution has a substantially lower average. There might be a way to weight the loss of longer context samples during training, such that the weighted train distribution matches the test distribution. The attention mechanism induces some amount of blurring between tokens, that increases based on the number of tokens in the context window. In general its impressive that attention can learn one set of weights that work for a variable context length. 

> I discovered yesterday that my PR's Data Generator includes **no** token 50256. This was an accident. However, since my PR appears to have lower loss than Ablation 3 in the above, I re-ran Ablation 3 but dropped the BOS token

From prior ablations [here](https://github.com/KellerJordan/modded-nanogpt/pull/117) I showed that dropping the bos token only had 0.003 increase in loss for the model where sparse attention gate was present. The attention gate causes less attention to be paid to the bos_token. As a result, it does not seem surprising that ablation 3 and ablation 3 w/ no BOS are relatively close.

ClassicLarry:
163!! 

args.train_max_seq_len = 2048 has some interesting theory behind it. For any given data distribution, fixed model, and loss function, one can compute the variance of the gradient over randomly sampled batches. Even though truncating sequence lengths introduces a bias by filtering tokens after 2048, it reduces the data correlation within a given batch, resulting in lower variance in the gradient and more stable training.

In this case there are competing factors of FA3 runtime, and the attention sliding window of 1400, but gradient variance as a function of max_seq_len is something that can be readily computed and inform modeling decisions.

varunneal:
@YouJiacheng Latest update uses `cu_seqlens_*` as per your proposal.

guilhermeleobas:
Hi @varunneal,

> The main idea of this record is to use [Flash Attention v3](https://github.com/Dao-AILab/flash-attention) instead of Flex Attention. The official version of this module is incompatible with torch.compile and causes graph breaks. However, a https://github.com/Dao-AILab/flash-attention/pull/1769 by @guilhermeleobas addresses this issue.

Glad to hear my work helped you. Just out of curiosity, was this work important to get the WR numbers?

varunneal:
@guilhermeleobas  Very important! Thank you very much for your contribution. Also interested in discussing more -- do reach out.

guilhermeleobas:
Cool. How important is flash attention with kvcache for you? More specifically, specifying a `num_splits` value?
https://github.com/Dao-AILab/flash-attention/blob/afc97c60f799e470886c154e3473df938f8fa93d/hopper/flash_attn_interface.py#L640-L671

Asking because torch will graph break today if `num_splits <= 0` and solving this requires redesigning part of the [heuristic](https://github.com/Dao-AILab/flash-attention/blob/afc97c60f799e470886c154e3473df938f8fa93d/csrc/flash_attn/flash_api.cpp#L263) used to compute the number of splits.

varunneal:
@guilhermeleobas Unimportant, since we do not care about inference speed.

PR #127, opened 2025-09-16
New WR 153.9s: Apply threading to preload data and extend final layer attention window for validation

ClassicLarry:
This PR builds on all recent WR improvements including PR #125 by @bernard24. This one adds:

- Start prefetching and indexing the next shard immediately. Since this occurs on the CPU, there is ample time to perform this during the GPU heavy workload, and we shouldn't be bottlenecking GPU activities on CPU data indexing. (1.5s)
- Only partially index the first shard before starting to train on it. Kickoff a parallel thread to finish indexing it, which gets picked up on the 5th step. (300ms)
- Extend the final layer attention window out to 20 for validation (no need to apply YaRN for this layer). If curious, some inspiration for this change came from: https://medium.com/@larry36d/formation-of-induction-heads-in-modded-nanogpt-5eb899de89e4. This dropped loss by roughly 0.001 and enabled -10 steps (1s) while still cleanly remaining under 3.28. 

(Exact runtimes will vary by GPU provider, mine is about 1s slower than I believe some GPU setups will get)

Validation:
```
import scipy.stats
import torch

accs = [3.2813, 3.2778, 3.2781, 3.2764, 3.277 , 3.2787, 3.2767, 3.2807,
       3.2769, 3.2774]

times = [153.869, 154.126, 153.756, 153.82 , 153.816, 154.014, 153.887,
       153.811, 154.015, 153.828]

print("p=%.4f" % scipy.stats.ttest_1samp(accs, 3.28, alternative="less").pvalue)
# p=0.0030
print("acc:", torch.std_mean(torch.tensor(accs)))
# acc: (tensor(0.0017), tensor(3.2781))

print("time:", torch.std_mean(torch.tensor(times)))
# time: (tensor(0.1180), tensor(153.8942))
```

PR #130, opened 2025-09-18
New WR 152.7s: Smear token embeddings 1 position forward, -15 steps

ClassicLarry:
This PR builds on all recent WR improvements including PR #127. From inspecting trained model weights, it was observed that multiple attention heads were consistently attending to the prior token. However, attention is a computationally inefficient way to attend to the prior token. This functionality is built-in below in a light-weight manner. The first 12 dimensions of the residual stream/embed are used to gate both the smear module and attention. Approximately, the model finds that (token + 0.07prior_token) is a more useful embedding representation than (token).

Note: This improvement is more marginal than the timing change would indicate. The prior WR had a mean loss of 3.2781. If I attempt to control for loss, the impact of this change appears closer to 5 steps based on testing. I think the main value of this PR is not the timing improvement, but the introduction of a new design space for optimizing the architecture: directly modeling close range information passing between tokens outside of attention.

```
self.smear_gate = CastedLinear(12, 1)
self.smear_gate.weight.detach().zero_()

x = self.embed(input_seq)
# smear token embed forward 1 position
smear_lambda = self.scalars[5 * len(self.blocks)]
smear_gate_out = smear_lambda * torch.sigmoid(self.smear_gate(x[1:, :self.smear_gate.weight.size(-1)]))
x = torch.cat([x[:1], x[1:] + smear_gate_out * x[:-1]])
x = x0 = norm(x[None])
```

Validation:
```
import scipy.stats
import torch

accs = [3.2781,3.2792,3.2765,3.2796,3.2803,3.2801,3.2787,3.2798,3.2787,3.2786]

times = [152.771,152.816,152.834,152.755,152.789,152.773,152.815,152.796,152.798,152.754]

# p=0.0084
# acc: (tensor(0.0011), tensor(3.2790))

# time: (tensor(0.0269), tensor(152.7901))
```

### DIFF
--- a/records/091825_Smear/18a1e5c7-947e-479d-bc3a-a57a61a98fc9.txt
+++ b/records/091825_Smear/18a1e5c7-947e-479d-bc3a-a57a61a98fc9.txt
@@ -1,0 +1,3089 @@
+import os
+import sys
+
+with open(sys.argv[0]) as f:
+    code = f.read()  # read the code of this file ASAP, for logging
+import copy
+import glob
+import math
+import threading
+import time
+import uuid
+from dataclasses import dataclass
+from itertools import accumulate
+from pathlib import Path
+
+os.environ["PYTORCH_CUDA_ALLOC_CONF"] = "expandable_segments:True"
+import torch
+
+torch.empty(
+    1, device="cuda", requires_grad=True
+).backward()  # prevents a bug on some systems
+import torch._dynamo as dynamo
+import torch.distributed as dist
+import torch.nn.functional as F
+
+# torch._inductor.config.coordinate_descent_tuning = True # we have banned this flag for new records because it causes compilation to take 30min
+import triton
+import triton.language as tl
+from flash_attn_interface import flash_attn_varlen_func
+from torch import Tensor, nn
+
+dynamo.config.recompile_limit = 64
+
+# -----------------------------------------------------------------------------
+# Custom operators: FP8 matmul by @YouJiacheng
+
+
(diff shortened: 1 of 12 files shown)


varunneal:
Great job. These sparse sigmoid gates seem really powerful. How much did you experiment with the inner dim of the gate? E.g. you've chosen `12` for this as well as the attention gate in #117

ClassicLarry:
> Great job. These sparse sigmoid gates seem really powerful. How much did you experiment with the inner dim of the gate? E.g. you've chosen `12` for this as well as the attention gate in #117

I only tried size 12. I also tried moving it to a different 12 dimensions than the 12 used by attention, but no improvement/worse.

Other things I tried: putting after norm, only applying to x instead of x0, gating on the prior token instead of current token, gating on both the prior token and current token, changing lr_multiple to 0.1 or 2.

trianxy:
Nice job, and thank you @ClassicLarry for adding this note
> Note: This improvement is more marginal than the timing change would indicate. The prior WR had a mean loss of 3.2781. If I attempt to control for loss, the impact of this change appears closer to 5 steps based on testing.

Such notes are extremely helpful in developing a better understanding of the architecture and the improvements. Sometimes people (especially in ML papers) leave out such ablation studies (or the results thereof), leaving me as the reader with less understanding (or even a mistaken conclusion) than I would have had otherwise.

YouJiacheng:
Last year, I tried simple learnable (not data-dependent) 1-token smearing (a.k.a. window=2 causal conv), following RWKV.
But it didn't achieve significant improvement so I dropped it.

PR #134, opened 2025-09-29
New WR: Polar express (-10 steps) + packaged Flash Attention quality of life improvement 

varunneal:
# New record 09/29/25

This submission reflects all recent WR changes up to [PR#133](https://github.com/KellerJordan/modded-nanogpt/pull/133).

The main improvement in this PR is using the [Polar Express](https://arxiv.org/pdf/2505.16932) sign method in Muon instead of Newton-Schulz. This paper was designed with reference to ModdedNanoGPT so it was very easy to implement, and I direct the reader to this paper directly for details. Using Polar Express, I've reduced the train steps by 10. 

The next change in this PR is packaging Flash Attention 3 via [Huggingface's Kernels](https://huggingface.co/docs/kernels/en/index). This does not impact timing but should increase ease of development for anyone working on this project.

## Timing and Validation

This PR improves the final training by 10 steps, with no change in the time per step.

```
import scipy.stats
import torch

losses = [3.2789, 3.2792, 3.2796, 3.2776, 3.2797, 3.2787, 3.2792]
times = [148.617, 148.580, 148.569, 148.653, 148.578, 148.542, 148.587]

print("p=%.4f" % scipy.stats.ttest_1samp(losses, 3.28, alternative="less").pvalue)
# p=0.0045

print("losses:", torch.std_mean(torch.tensor(losses)))
# losses: (std=0.0007057, mean=3.2789857)

print("time:", torch.std_mean(torch.tensor(times)))
# time: (std=0.0358076, mean=148.5894318)
```

You may notice that this PR shows a 0.2 second mean *increase* in timing over the result in PR#133. However, that PR was timed on very fast machine. To demonstrate that this PR accurately represents a decrease in train time, I timed PR#133 on the same machine as above:

```
import scipy.stats
import torch

times = [149.714, 149.676, 149.659, 149.716, 149.649, 149.569, 149.521]

print("time:", torch.std_mean(torch.tensor(times)))
# time: (std=0.0732, mean=149.6434)
```

Therefore, I believe that this PR represents at least a 1 second improvement. 

Thank you to Prime Intellect for compute credits, which made this PR possible. 

## Polar Express

All credit to the original authors (Noah Amsel, David Persson, Christopher Musco, Robert M. Gower) for discovery and implementation of this method. I adapted their code from https://github.com/NoahAmsel/PolarExpress/tree/main.

I found optimal parameters with
- `num_iters=5`: each iteration adds about a second to train time
[...]

PR #136, opened 2025-09-29
New record 146.8s 09/30/25 (-0.6s): CustomBatching, only update Adam Params every other step

ClassicLarry:
This submission reflects all recent WR changes up to [PR#134](https://github.com/KellerJordan/modded-nanogpt/pull/134).

The main contribution of this PR is to introduce the concept of variable batch size by parameter group, by having different gradient accumulation strategies for different param groups. I drop the batch size by 1/3, and then allow gradients to accumulate in Adam params over 2 steps before updating, whereas Muon params update every step.
Also, for a minor improvement, I add a cooldown to the momentum term in Muon for the last 50 steps.

* num_iterations: 1630->2380
* cooldown_frac: 0.5->0.4
* adam beta1: 0.8->0.7

The mean loss of 3.2765 is well below the 3.28 cutoff. There may be some hyperparameter tuning that can be performed, even as simple as decreasing the step count. I only tried 1 batch size, 1 cooldown frac, and only minimally looked at parameter learning rates or momentum terms. There is a consistent hiccup of 300ms around step 2385 that may be fixable. Overall I encourage hyperparameter tuning as that was left out of scope for this PR. This investigation into batch size was partially motivated by conversations with @varunneal.

```
# only step Adam every other step
if step%2==0:
    optimizer2.step()
    optimizer2.zero_grad(set_to_none=True)
else:
    for opt in optimizers:
        opt.step()
    # null the gradients
    model.zero_grad(set_to_none=True)
```

## Timing and Validation
```
import scipy.stats
import torch

losses = [3.2745, 3.2747, 3.2771, 3.2794, 3.2767]
times = [146.933, 146.893, 146.943, 146.553, 146.739]

# p=0.0086

# losses: (tensor(0.0020), tensor(3.2765))

# time: (tensor(0.1664), tensor(146.8122))
```

retiming prior record: 147.4: [147.451, 147.336, 147.508]


Gusarich:
> The mean loss of 3.2765 is well below the 3.28 cutoff.

You could easily decrease steps by 15-20 and cut off a whole second while still being under 3.28 loss!

ClassicLarry:
> > The mean loss of 3.2765 is well below the 3.28 cutoff.
> 
> You could easily decrease steps by 15-20 and cut off a whole second while still being under 3.28 loss!

I am done testing for a while, but yes the next couple seconds should fall pretty quickly.

varunneal:
@ClassicLarry I believe this record also introduced the cooldown phase for Muon. What was the motivation for it in this PR?

ClassicLarry:
> @ClassicLarry I believe this record also introduced the cooldown phase for Muon. What was the motivation for it in this PR?

I don't know how accurate this is, but I have a mental image of the model during training moving down the side of a snowy mountain, and the final low lr regime is the model committing to a local crevasse. When taking really small steps in this crevasse, I don't want to be influenced by old gradients, as I want to have the agility to move with the angle of the crevasse. That was the theoretical motivation. 

The second factor is that during end-of-training optimizations is really easy to tune because you can save a couple model runs right near the end, load them into memory and then just tune a parameter on 50 steps.

So I tuned this on an older Colab model and saw 0.001 loss improvement. Afterwards, I added the much more radical changes on batch size and did not retune the momentum cooldown.

PR #138, opened 2025-10-01
New Medium WR: -27 seconds via Layer Reuse; includes #137

snimu:
I simply added the output of layer 11 (the 12'th layer) to the output of the final layer (layer 15, or the 16'th layer), in a weighted sum. The weights are scalars and are learned. This just means that right before applying the language head, I do this:

```python
skip_lambdas = self.scalars[-2:]
x = norm(x) * skip_lambdas[0] + norm(skip_connections[11]) * skip_lambdas[1]
```

Where the `skip_connections` contain the output latents of each layer, at the corresponding position.

This allowed me to reduce the step count to 5560 (-30 steps)

These are the resulting final validation losses over 10 runs:

```python
[2.918849, 2.917008, 2.919394, 2.91776, 2.919522, 2.919678, 2.918775, 2.918111, 2.919395, 2.917938]
```

And here are simple stats about the final validation loss over these 10 runs:

- Mean: 2.9186
- Std: 0.0008985
- P-value: 0.0005035724941803675

The p-value is very low, so after only 10 runs we can be sure that the target loss is reached.

Now here are the corresponding run times:

```python
[1378.689, 1378.504, 1378.622, 1379.238, 1379.0, 1379.444, 1378.426, 1379.337, 1378.254, 1379.127]
```

And here are some simple stats about the times:

- Mean: 1378.8641 seconds ~= 22.98 minutes
- Std: 0.4177 seconds

The previous record's time is 1405.698 seconds, so this is a reduction of almost 27 seconds. This seems like a lot for 30 steps, and may thus partially be an artifact of using different GPUs (though I have to say that a p-value this low after 10 runs indicates that I could maybe have reduced the step count even more).

The reason why I chose layer 11 is that I performed the following ablations over the layers:

image

And in that single run, layer 11 lead to the minimum loss.

### DIFF
--- a/records/100124_GPT2MediumLayerReuse/000_9dbc0505-5fc3-4a3c-b520-88834bbf40b3.txt
+++ b/records/100124_GPT2MediumLayerReuse/000_9dbc0505-5fc3-4a3c-b520-88834bbf40b3.txt
@@ -1,0 +1,6612 @@
+import os
+import sys
+
+with open(sys.argv[0]) as f:
+    code = f.read()  # read the code of this file ASAP, for logging
+import copy
+import time
+import uuid
+from dataclasses import dataclass
+from functools import lru_cache
+from pathlib import Path
+
+os.environ["PYTORCH_ALLOC_CONF"] = "expandable_segments:True"
+import torch
+
+torch.empty(
+    1, device="cuda", requires_grad=True
+).backward()  # prevents a bug on some systems
+import torch.distributed as dist
+import torch.nn.functional as F
+from torch import nn, Tensor
+
+# use of FlexAttention contributed by @KoszarskyB
+from torch.nn.attention.flex_attention import BlockMask, flex_attention
+
+
+torch._inductor.config.coordinate_descent_tuning = (
+    True  # we allow this flag for medium track
+)
+# torch._dynamo.config.compiled_autograd = True
+
+
+class Snoo:
+    """
+    @DominikKallusky, @vishal9-team, @vinaysrao
+
+    Sparse Nesterov Outer Optimizer (Snoo) is a momentum-based wrapper to any optimizer that can
+    improve the stability and smoothness of the optimization process and thus the quality
+    of large language models (LLM) and other models. Snoo implicitly adds temporal regularization
+    to the parameters, thus smoothing the training trajectory and instilling a bias towards flatter
+    minima and lower parameter norms. Snoo is computationally efficient, incurring minimal overhead
+    in compute and moderate memory usage.
+    """
+
+    @torch.no_grad()
+    def __init__(self, model: nn.Module, lr: float, momentum: float, k: int) -> None:
+        self.model = model
+        self.lr = lr
+        self.momentum = momentum
+        self.k = k
+        self.current_step = 0
+        self.outer_buf = [p.clone() for p in model.parameters()]
+        self.model_params = list(self.model.parameters())
+        self.optimizer = torch.optim.SGD(
+            self.model.parameters(),
+            lr=lr,
+            momentum=momentum,
+            nesterov=True,
+            fused=True,
+        )
+
+    @torch.no_grad()
+    def step(
+        self,
+    ) -> None:
+        if self.current_step % self.k == 0:
+            for p_new, p_old in zip(self.model_params, self.outer_buf):
+                p_new.grad = p_old.data - p_new.data
+                p_new.copy_(p_old, non_blocking=True)
+
+            self.optimizer.step()
+
+            for p_new, p_old in zip(self.model_params, self.outer_buf):
+                p_old.copy_(p_new, non_blocking=True)
+        self.current_step += 1
+
+    def state_dict(self):
+        state_dict = {
+            "current_step": self.current_step,
+            "lr": self.lr,
+            "momentum": self.momentum,
+            "k": self.k,
+            "outer_buf": [p.clone() for p in self.outer_buf],
+            "optimizer_state_dict": self.optimizer.state_dict(),
+        }
+        return state_dict
+
+    def load_state_dict(self, state_dict):
+        self.current_step = state_dict["current_step"]
+        self.lr = state_dict["lr"]
+        self.momentum = state_dict["momentum"]
+        self.k = state_dict["k"]
+        for p_src, p_dst in zip(state_dict["outer_buf"], self.outer_buf):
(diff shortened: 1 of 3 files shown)


ClassicLarry:
Very cool idea. That is interesting that this is done after the norm. There may be some application here to the short track, even just applying a scalar weight after the norm. I'd be curious where these two scalars land at the end of training.

Its challenging to tell what code was updated exactly, since the update commit is mixed in with a refactor of the entire file. I'm reading this is as at most a 7 second improvement (30 steps), with a substantial mean loss drop to 2.9186. The 27s is misleading because the majority of that is from the faster GPU, which should be controlled for. If the new skip adds over 0.5% runtime, then this isn't a new record. (my guess is this is a new record, and its easily an improvement with the loss drop)

snimu:
Those are good points.

- I'll redo the baseline tomorrow for a few runs and compare the plots. I did compare the two before the step reduction and my method was very clearly faster, so I also think that it's a record, but I'll run the tests again
- As for the scalars, I should absolutely log them. Will do and report.
- I'll also try to norm-then-sum-then-norm again, to see if a scalar weight at the output makes a difference (I doubt it because the language head should be able to make up for it).
- I also have plans for adding in multiple previous activations (I want to perform pretty thorough ablations and write a blog post, so the results will be interesting no matter what happens)

About the commit, yeah I should have done two commits, sorry. It's really barely any change at all; just the two lines I've shown in the PR description and creating the two lambdas, initialized to 1.0 for x and 0.0 for the skip connection.

Thanks for the feedback:)

snimu:
Oh and the norms before the sum are of course needed because the activation magnitude rises throughout the layers, so the scalars would be meaningless without the norm and I also doubt the model would learn enough.

snimu:
Here's my initial plot of timings, but that's only for one run each:

image

(that the baseline from #137 doesn't cross 2.92 here is probably just random chance, but even if we bent the curve down so that it does, my run gets there faster)

ClassicLarry:
> I also have plans for adding in multiple previous activations (I want to perform pretty thorough ablations and write a blog post, so the results will be interesting no matter what happens)

Will be interesting to get to the root cause of the improvement.

> Oh and the norms before the sum are of course needed because the activation magnitude rises throughout the layers, so the scalars would be meaningless without the norm and I also doubt the model would learn enough.

I agree that norm before is probably better, but the scalar could learn to compensate for the magnitude so I wouldn't consider it meaningless.

>I'll also try to norm-then-sum-then-norm again, to see if a scalar weight at the output makes a difference (I doubt it because the language head should be able to make up for it).

Probably, but there are a lot of times where the model 'should' be able to learn something that it ends up learning slower and getting higher loss. As an example, on the small track the logits were scaled down 27x and lr scaled up to match, but this increased loss because the lm_head weights ended up 27x bigger which didn't do well with the tuned second order effects of Adam.

One direction you could take this: When an MLP or ATTN layer adds an output to the residual stream, that output has to serve two functions simultaneously. It give better context to downstream layers, and it also is directly contributing to the final prediction. This means that if a layers output is only useful for downstream context, a later layer has to reverse out the contribution. These functions could be separated by making the final prediction reapply a lambda to each mlp_out and attn_out. Idea probably wont work in practice, but just thinking out loud. Thinking of your update in terms of lambda1*norm(attn0+mlp0+...mlp11) + lambda2*norm(attn0+mlp0+...mlp15).

snimu:
A few early results:

- The x-lambda tends to be around 1 (0.7 - 1.3), the skip-lambda is always negative (~ -0.3). So your theory that this technique allows for a separation of concerns is correct: layer 11 provides context for layer 12, but actively removing it from the output latent improves prediction; thus, the presence of layer-11-latents at the output is harmful (I'll cite you in my article!). Very curious now how the lambdas will look when I add multiple latents.
- My run is faster than #137 but only by a little bit; since I have such a generous p-value I will reduce step count by 5 and try again. If that works out, I'll close this PR and open another one with a clean commit history, good description, direct timing comparison to #137, and clearer record

snimu:
Another thing: norming again after the sum doesn't improve final performance; if anything, it very slightly hurts it. So maybe, just maybe, adding a learned lambda alone could already improve performance? I'm not sure, the difference is very small and over a sample size of one, so it's probably just irrelevant.

ClassicLarry:
On the short track I applied a sparse gate for attention and token smearing. Adding a sparse gate here may be helpful so that lambda depends on the local context and not just the layer number. I’ll prob test that and some other extensions on the short track later.

snimu:
Cool :) My proper ablations will have to wait because the ACL deadline is on Monday and I cannot get the GPUs I want (SMX5 Hyperstack 8xH100, they are by far the most consistent and fastest that I have tested). I'll also look into the PRs you've mentioned sometimes next week.

ClassicLarry:
I am considering a 12 lambda approach on the short track where each weight is setup to back out the contribution made by each layer. Tested once on older version of short track for now and results were very promising.
```
# init:
backout = torch.zeros_like(x)

# for each layer:
x_out = self.blocks[i](x, x0, lambdas[i], attn_args)
backout += backout_lambda[i]*(x_out-x)
x = x_out

#at end:
x = x - backout
x = norm(x)
```

Final weights:
[ -0.6019, 0.5400, 0.4613, 0.4364, 0.3429, 0.2675, 0.3030, 0.2023,
        0.3761,  -0.0741,  -0.2164,  -0.2905]

The model learns to partially back out the contributions from layers 2-9 and upweight 1,10-12. The approach you tested is somewhat similar to having a single shared backout lambda across the first 2/3 layers. The average of the first 2/3 lambdas here is 0.31, which matches your -0.3 (I flip the sign convention).

snimu:
Oh that's very clever.

My plan is now the following (and I'd appreciate feedback on it):

- close this PR
- make a new, clean PR with my original technique (I did get some GPUs); I should be able to do this fairly soon, and it would be a good first waypoint (looks like ~8.5 to 9 seconds in time reduction currently)
- run ablations about simply adding multiple layer outputs (up to all of them) to the final output in a learned weighted sum (this is well underway, starting very soon)
- then do ablations with the backout technique you've shown (with credits to you obviously)

ClassicLarry:
sounds good! I am finding on the small track a single lambda on layer 8 similar to your setup performs best in terms of compute/accuracy tradeoff. Will raise a PR later today.

snimu:
Re-done in #139

PR #141, opened 2025-10-13
Implement NorMuon

zichongli5:
### 🚀 Add NorMuon Optimizer (Neuron-wise Normalized Muon)

This PR adds **[NorMuon](https://arxiv.org/pdf/2510.05491)**, a variant of the Muon optimizer that introduces **neuron-wise normalization** to improve stability and convergence efficiency.

#### Results
Built on the current master branch (excluding recent PRs), NorMuon reaches the **3.28 loss cutoff in 1700 steps**, 50 steps faster than Muon with identical per-step time.
 With 1700 steps schedule, Muon achieves a validation loss of **3.2859**, while NorMuon achieves **3.2784 / 3.2795 / 3.2793** across three runs.

#### Hyperparameters
We keep the same hyperparameters and have not tuned them for NorMuon. The performance might be further improved with tuning.

#### Notes
This implementation includes two slight modifications from the paper version:

1. The update is rescaled to match the norm of the pre-normalized update.
2. In the algorithm from the paper (line 7), we average the second-order momentum across columns. In this implementation we average across columns if m ≥ n and across rows otherwise, which performs slightly better.

### DIFF
--- a/records/101225_NorMuon/115b1530-f861-4365-be25-72dae24798ab.txt
+++ b/records/101225_NorMuon/115b1530-f861-4365-be25-72dae24798ab.txt
@@ -1,0 +1,2512 @@
+import os
+import sys
+with open(sys.argv[0]) as f:
+    code = f.read() # read the code of this file ASAP, for logging
+import uuid
+import time
+import copy
+import glob
+from dataclasses import dataclass
+from functools import lru_cache, partial # Added partial for hook registration
+from pathlib import Path
+
+os.environ["PYTORCH_CUDA_ALLOC_CONF"] = "expandable_segments:True"
+import torch
+torch.empty(1, device="cuda", requires_grad=True).backward() # prevents a bug on some systems
+from torch import Tensor, nn
+import torch.nn.functional as F
+import torch.distributed as dist
+# use of FlexAttention contributed by @KoszarskyB
+from torch.nn.attention.flex_attention import BlockMask, flex_attention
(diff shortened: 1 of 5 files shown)


ClassicLarry:
Awesome, this improvement looks quite substantial! The code update also looks very compact for the size of impact. I am going to merge in the other PRs this week on the 124M track, which are now 25s faster than this one on its own. However, I'm quite confident if these same changes get applied to the latest PR this will show a notable improvement, with maybe some minor tuning to account for change in batch size. Please see rule 2 in readme for details on p value for submission.

varunneal:
@zichongli5 it looks like this implementation is pretty similar to adafactor, but instead of computing a full 2d mesh, it just computes it along one dimension (e.g. either along rows or cols). I'm wondering
- why not apply the second moment along both rows *and* cols?
- why not apply the second moment prior to the zero power function?

zichongli5:
> @zichongli5 it looks like this implementation is pretty similar to adafactor, but instead of computing a full 2d mesh, it just computes it along one dimension (e.g. either along rows or cols). I'm wondering
> 
> * why not apply the second moment along both rows _and_ cols?
> * why not apply the second moment prior to the zero power function?

Yes — this follows the similar idea as adafactor/adammini.

For the first question: That’s a valid extension, though I kept it 1-D for simplicity and minimal state; applying both might give a small gain.

For the second question: We tested applying second moment before ns5 in the paper (see Fig. 3) and found very similar results.

ClassicLarry:
Closing as #144, an updated version of this PR, has been merged.

PR #163, opened 2025-11-30
New Record: Batch size schedule (-2 seconds)

varunneal:
This record implements a batch size schedule.

## Timing and Validation

This record has 65 fewer steps at slightly higher average step time than the current record.

```
import scipy.stats
import torch

losses = [3.2772, 3.2781, 3.2780, 3.2784, 3.2773, 3.2806, 3.2799, 3.2765]
times = [132.259, 132.152, 132.194, 132.179, 132.249, 132.190, 132.217, 132.275]

# p=0.0045

# losses: (std=0.0014, mean=3.2782)

# time: (std=0.0431, mean=132.2144)
```

```
import scipy.stats
import torch

times = [134.069, 134.013, 134.044]

# time: (std=0.00281, mean=134.042)
```

These timings show an improvement of $\approx 1.8$ seconds.

## Batch Size Schedule

Choosing a batch size involves a tradeoff:
1) Token efficiency. Smaller batch sizes are more token-efficient. For a fixed token budget, once the batch size is above some threshold, increasing it further tends to hurt final model performance. Beyond this point, the gradient signal from a single batch is saturated, so larger batches just waste tokens.
2) Token speed: Training at higher batch sizes is faster *per-token* than training with a lower batch size for many steps. This is because GPUs parallelize over the batch dimension, DDP is easy, and every step incurs overhead (optimizers, comms, etc).

A critical batch size balances both of these considerations: low enough to be token-efficient, but high enough to be speed-efficient.

Batch size can often be safely increased over the course of training. Conceptually, this may be because early training focuses on common patterns, so even small batches provide strong gradient signals. Later training involves learning rarer patterns. These sparse signals remain noisy even at high batch sizes, so the batch size can be increased without saturating the gradient. Notably, two strong models trained with Muon, Kimi K2 and GLM 4.5, both increase batch size mid-training.

Batch size schedules have been unsuccessfully tried in the speedrun for some time now. It is a multifaceted problem, requiring consideration of token budgets, optimal learning rates, and token efficiency.

In order to simplify the problem, I assumed that the total token budget should be approximately the same, but the batch size should be linearly ramped over three phrases of training (this is also for convenience -- to match the window size schedule). Within these constraints, I only needed to optimize (a) the slope of the linear ramp and (b) the learning rate schedule.

There are certainly much richer sweeps of hyperparameter sweep that could drive down the results in this record.

If you are interested in some experiments I conducted for this record, as well as surronding literature, please do read Section 2 of my recent blog post [Muon in Modded NanoGPT](https://varunneal.github.io/essays/muon).

Additionally, I have attached logs with validation every 5 steps in this repo for more fine-grained analysis of the batch size schedule. Here is a before/after plot:

## PyTorch version

The previous nightly (Torch 2.10, 09/26/2025) used for most records has expired. Unfortunately, both Torch 2.9.0 and Torch 2.9.1 seem to slow down the record. As such, I traveled significantly forward to nightly 12/04/2025, which seems to be very slightly faster than the previous nightly.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -3,3 +3,4 @@ tqdm
 torch
 huggingface-hub
 kernels
+setuptools


varunneal:
I noticed while trying to find an optimal learning rate that quite a few LRs (both high and low) were causing NaNs. I didn't face that issue with the current schedule but I may try to check if this has to do with the polar express coefficients. Maybe they can be adjusted before merging

ClassicLarry:
Awesome this will be great addition. Getting the batch size schedule to actually work out and improve the record is not as easy as it may look!

Since the gradients are summed within a batch during training, increasing the batch size will have the side-effect of increasing the gradient magnitude by 2x then 1.5x, which will cause the momentum buffer to clear out faster after the window size schedule change. This seems good to me, but potentially that 'momentum buffer clear out rate' could be tuned separately by rescaling the buffer magnitudes at 1/3 and 2/3 of training.

A loss curve here, with validation loss sampled at +-5 steps of the transition points would be interesting. I'm curious if it spikes at all during the transition.

varunneal:
@ClassicLarry updated PR. Should be good to merge. The earlier NaNing seemed to be because of Adam's low tolerance for high learning rates (esp at high batch sizes). Not really an issue, just an artifact of testing out learning rates that were too high. I've changed the `get_lr` logic a bit to use exponentiation instead.

ClassicLarry:
thanks for finding a new good nightly. Chart is interesting. Loss spike at 1/3 indicates the model has to recalibrate itself somewhat during the batch size transition. The minor loss bump around step 200 is peculiar. Makes me wonder if its a bad training batch- maybe there is something going on here where gradient clipping would be useful.

PR #164, opened 2025-12-09
DistAdam was never warmed up in the warm-up loop. Fix.

shenberg:
DistAdam.should_sync was never set to True in the warm-up loop so `step()` and `_sync_gradient` didn't get to see realistic work. Seems to shave some run-time off.
In general, making the warm-up more like the actual loop because I've been getting weird intermittent long steps and I'd rather have Inductor not work at all during the training loop even if it's inside the validation step because it does influence the heap etc. Trying to eliminate as much variance as possible.

Found this by setting `TORCH_LOGS=recompiles`


shenberg:
Encountered some problems when re-doing this on the newer code

shenberg:
See #165

PR #181, opened 2025-12-27
New Record: Logit Rescale (-40 steps, -2.9s)

ClassicLarry:
Old code:
```
logits = self.lm_head(x)
logits = 30 * torch.sigmoid(logits / 7.5)
```

New code:
```
logits = self.lm_head(x)
logits = 23 * torch.sigmoid((logits+5) / 7.5)
```

image

The activation output after logits=self.lm_head(x) is:

image

after applying the sigmoid and range rescaling the activation output is:

image

Also adding some training refactoring. Happy to merge in anything that improves code clarity.
I would expect only 40*.062 = 2.5s from this record. The remaining 0.4s may be from a more thorough warmup.

## Timing and Validation
```
import scipy.stats
import torch

losses = [3.2794, 3.2777, 3.2793, 3.2799, 3.2784, 3.277, 3.2778, 3.2803, 3.2793]
times = [116.405, 116.313, 116.483, 116.359, 116.521, 116.3, 116.536, 116.402, 116.36]

# p=0.0058

# losses: (tensor(0.0011), tensor(3.2788))

# time: (tensor(0.0868), tensor(116.4088))
```

retiming prior record matches official time: 119.3: [119.432, 119.138, 119.329]


chrisjmccormick:
Yeah, I like the refactoring. The training loop looks a lot cleaner.
I like the `activate_hooks` and `_is_active_step` functions, those are a nice approach.

chrisjmccormick:
Congrats on the record here, that's quite a big jump! 

What's the theory behind the logits rescaling?

ClassicLarry:
> Congrats on the record here, that's quite a big jump!
> 
> What's the theory behind the logits rescaling?

Tbh there's a large part here of "seeing a number and wondering what happens if I change it". If I think about it after the fact, the 30->23 change is limiting how certain the model can be in a prediction. Before, a model could assign e^30=1e13 times higher probability to one logit over another. For a more sophisticated model this level of precision might be useful. However at 3.28 loss it appears counter productive.

For instance on a logit scale of 0-23 if the preferred token gets the max of 23 and the other 50,000 tokens get 5, then e^23/(e^23+e^5*50000) = 0.999, gives 99.9% prediction for the preferred token. Perhaps we don't really ever need to be more certain than that.

The charts above show that the baseline logit activation distribution after x=lm_head(x) is not centered at zero. Since our data is already not symmetric around zero, I figured it could be useful to test asymmetric rescaling functions. In this case it looks like the new activation distribution has only the promising logits set to positive values, which is typically a small number of tokens.

ClassicLarry:
> Yeah, I like the refactoring. The training loop looks a lot cleaner. I like the `activate_hooks` and `_is_active_step` functions, those are a nice approach.

thanks! Unfortunately I accidently dropped the call to start_transition() that should live in advance_schedule() to freeze the scalars. Can probably get corrected in the next PR by whoever raises it.

PR #183, opened 2025-12-28
SwitchBank medium-track record

HaileyStorm:
# SwitchBank (Track 2 Medium)

This run uses the SwitchBank model instead of the baseline GPT-2 Medium stack. SwitchBank is a "horizontal MoE" architecture designed by me (Hailey Collet):
each transformer layer routes tokens into a shared FFN expert bank, while the attention stack remains per-layer.
This gives independent control over attention depth vs FFN capacity, and allows many layers to share the same expert weights.

## What is different vs train_gpt_medium.py
- Shared FFN bank across layers with per-layer routers (shared experts, independent routing).
- Top-k hard routing with optional Gumbel noise; auxiliary loss mixes load/importance CV2 and entropy gap.
- Router features include token norms plus optional forward/reverse EMA context and document-aware flags.
- Optional router adapters (per-layer/per-expert scale/bias) with lazy init and pruning hooks.
- Router temperature + logit-cap schedules anchored to second-expert activation.
- TurboMuon tweak.
- Code is spread over multiple files: train_switch_bank.py and the switch_bank package. The former prints all of the code from both in the reproducible log.

## Routing/adapter specifics (SwitchBank)
- Each layer computes gates into the same expert bank; outputs are scaled by gate probabilities.
- Adapters can be enabled on routers and optionally frozen late in training.
- Routing supports temperature and logit-cap schedules, optional Gumbel exploration, and per-layer boost shapes.
- Bank-level metrics track load/usage/entropy and a composite router health score.

## Observations from nearly 1000 runs / ablations
- Reverse EMA features were much more effective than forward EMA (regardless) or both (by wall time).
- DeepSeek-style routing did not work here after extensive tuning. Best attempt was sigmoid-only routing
  (bias/update variants were worse); it still significantly underperformed by loss vs step and time. Router health did improve with sigmoid.
- TurboMuon was slightly better by step and slightly worse by time; router health improved, so it stayed enabled.
- Moving from 16 to 28 layers was a big win. It matched ~3.03 loss (best value at the time) in similar wall-clock time but in fewer steps,
  and the curve was steeper. With more time it reached the target loss and below. This highlights the value of
  SwitchBank: 1.75x layers for only ~1.16x parameters (~38.9M extra), and the ability to scale attention depth
  separately from FFN bank size.
- There remains lots of room for optimization, particularly if unconcerned with router health.
- There also remains lots of room for lower loss with the current model.

## Record facts (**preliminary** and based on 3 runs, pending official verification)
- 1750 steps vs the current record's 5960.
- 282.4M parameters vs 454.5M in `train_gpt_medium.py` (GPT-2 Medium reference ~350M).
- 13.86 minutes vs 24.07 (24.37 as I tested / on this record's hardware) - the largest absolute/percentage drop for small or medium track since their initial runs.
- 167.2M vs 3.12B tokens!
- Mean loss 2.896 vs 2.919 previous record / 2.92 target.
- Given I was able to test with few runs, I am OK if the step count needs to be increased some during confirmation.

## Notes
- Training/validation streams stay in order with the same FineWeb binaries as the baseline.
- Final validation uses the first 10,485,760 tokens of the validation set, per rules. Previous validation steps (optionally) use fewer tokens.
image
image
image


xTimeCrystal:
Reverse EMA leaks next token info, am skeptical of casual masking

xTimeCrystal:
More specifically the issue lies in this function:

_ema_reverse_since_doc_start

It allows token i to look at information from token i+1 which is against the point of casual masking and the model can just cheat by learning how to separate current token and next token information from the ema.

HaileyStorm:
> More specifically the issue lies in this function:
> 
> _ema_reverse_since_doc_start
> 
> It allows token i to look at information from token i+1 which is against the point of casual masking and the model can just cheat by learning how to separate current token and next token information from the ema.

Thanks for the fast response!

You're right that it leaks information **for the first `ema_window_size_rev` tokens of each document** = 384 tokens! It's causal after that (also not a problem while there's only one expert), and of course it's "only" part of an EMA, so the information is imperfect. I definitely missed this issue, and I will retest tonight/tomorrow. Hopefully the effect isn't *too* serious!

Thank you!

HaileyStorm:
Oh, also in theory this should only impact the routing? It's obviously not good, but... another indication it hopefully won't be a *huge* issue.

ClassicLarry:
Marking as draft until validation losses reflect updates, to reduce confusion for people trying to build on top of the latest record. I'm not sure if this will still be a record after the fixes but the architecture is certainly innovative.

HaileyStorm:
Ah, good call, thank you. 
Will update!

HaileyStorm:
I have received some credits (shout out when I report results).

Haven't felt great last couple days and I'm pretty busy today, but I'm determined to give officially beating the record my all, and prove the architecture *works* regardless!

HaileyStorm:
This dream is dead, at least for now: https://x.com/i/status/2010791813964296558

PR #186, opened 2025-12-29
New Record: Value Embed and Skip Gates (-35 steps, -1.3s)

ClassicLarry:
Updates in PR:
* Add gates to value embeds, using same structure as attention gates (first 12 dims of residual stream)
* Add gate to skip connection, using same structure as smear gate
* Update w_s from 1.5/448 to 1.6/448
* Add wd_mul 5 to value embeds, wd_mul 150x to embed and lm_head.
* Drop 35 steps

When I tested re-adding the frozen scalars for 40 steps (that got removed last PR) I got worse loss, so I kept it off for now. Maybe worthwhile to re-add for just the first window transition, not sure. The biggest change here is the value_embed gates. Loss is around 3.2830 without them. The skip_gate is pretty marginal, but it only costs ~300ms to add and seems to slightly pay for itself. I haven't heavily ablated the weight decay parameters, there's a chance one of those is actually detrimental. At some point we should update the notation of how weight decay is calculated to be lr/init_lr * lr instead of lr^2. lr^2 is making the wd_mul rather unintuitive.

I also tested adding gates for the x0 contributions but saw no improvement and higher runtime.

Motivation for w_s change was I started seeing very high variance results at some point during testing and recalled that the fp8_scales was how variance was previously resolved. Recent changes have increased the lm_head weights by roughly 10%, which aligns with the 1.5 -> 1.6 increase of w_s.

image

(notation in graph: old=record back with prior tuned fp8 scales, baseline=this PR. only plotting odd steps so step count is halved)

## Timing and Validation
```
import scipy.stats
import torch

losses = [3.279, 3.2795, 3.2774, 3.2779, 3.2784, 3.2792, 3.28]
times = [114.508, 114.464, 114.373, 114.343, 114.22, 114.401, 114.369]

# p=0.0061

# losses: (tensor(0.0009), tensor(3.2788))

# time: (tensor(0.0922), tensor(114.3826))
```

retiming prior record (faster machine this time): 115.686 [115.669, 115.717, 115.674]

If no changes, will merge at 115.1s to be consistent with 1.3s improvement.

I tried a large number of variants of gradient clipping and some custom ones, nothing I tested showed an improvement. I also tried a custom relu where non active neurons simulate the gradient as if they were at 0.01, performed much worse. Also tried a version of attention where the keys retrieve value deltas instead of values, performed worse.

### DIFF
--- a/records/track_1_short/2025-12-29_VeSkipGates/2851d7dc-d6a5-4e74-8623-57031425db16.txt
+++ b/records/track_1_short/2025-12-29_VeSkipGates/2851d7dc-d6a5-4e74-8623-57031425db16.txt
@@ -1,0 +1,3691 @@
+import os
+import sys
+
+with open(sys.argv[0]) as f:
+    code = f.read()  # read the code of this file ASAP, for logging
+import copy
+import glob
+import math
+import threading
+import time
+import uuid
+from dataclasses import dataclass
+from collections import defaultdict
+from itertools import accumulate
+from pathlib import Path
+import gc
+
+os.environ["PYTORCH_ALLOC_CONF"] = "expandable_segments:True"
+import torch
+
+torch.empty(
+    1, device=f"cuda:{os.environ['LOCAL_RANK']}", requires_grad=True
+).backward()  # prevents a bug on some systems
+import torch._dynamo as dynamo
+import torch.distributed as dist
+import torch.nn.functional as F
+
+# torch._inductor.config.coordinate_descent_tuning = True # we have banned this flag for new records because it causes compilation to take 30min
+import triton
+import triton.language as tl
+from kernels import get_kernel
+from torch import Tensor, nn
+
+dynamo.config.recompile_limit = 64
+
+
+# -----------------------------------------------------------------------------
+# Custom operators: FP8 matmul by @YouJiacheng
+
+@torch.library.custom_op("nanogpt::mm", mutates_args=())
+def mm_op(x: Tensor, w: Tensor, x_s: float, w_s: float, grad_s: float) -> tuple[Tensor, Tensor, Tensor]:
+    @torch.compile
+    def impl(x: Tensor, w: Tensor):
+        assert x.is_contiguous() and w.is_contiguous()
+        x_f8 = x.div(x_s).to(torch.float8_e4m3fn)
+        w_f8 = w.div(w_s).to(torch.float8_e4m3fn)
+        out = torch._scaled_mm(
+            x_f8,
+            w_f8.T,
+            out_dtype=torch.bfloat16,
+            scale_a=x.new_tensor(x_s, dtype=torch.float32),
+            scale_b=x.new_tensor(w_s, dtype=torch.float32),
+            use_fast_accum=True,
+        )
+        return out, x_f8, w_f8
+
+    return impl(x, w)
+
+@mm_op.register_fake
+def _(x: Tensor, w: Tensor, *_):
+    assert x.ndim == w.ndim == 2
+    assert x.shape[1] == w.shape[1]
+    assert x.device == w.device
+    assert x.is_contiguous() and w.is_contiguous()
+    return x @ w.T, x.to(torch.float8_e4m3fn), w.to(torch.float8_e4m3fn)
+
+@torch.library.custom_op("nanogpt::mm_backward", mutates_args=())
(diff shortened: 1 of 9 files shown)


chrisjmccormick:
Nice! 

I was just starting to look at increasing the weight decay on the value embeddings. Their weight values are big!

I looked into the scalar freezing yesterday, there's a couple changes to make beyond just calling `start_transition`.

Step 0 shouldn't be counted as a transition:

```python
    def get_transition_steps(self):
        transition_steps = []
        ws_short, ws_long = get_ws(0)
        for step in range(1, args.num_iterations):
            ws_short, new_ws_long = get_ws(step)
            if new_ws_long != ws_long and step != 0:   # <-- exclude step 0
                transition_steps.append(step)
                ws_long = new_ws_long
        return transition_steps

    def advance_schedule(self, step: int):
        self.ws_short, new_ws_long = get_ws(step)
        if new_ws_long != self.ws_long and step != 0:   # <-- exclude step 0
            self.model.yarn.apply(self.ws_long, new_ws_long)
            self.start_transition()
```

And the freeze timer needs to be set to `0` in reset:

```python
    def reset(self, state=None):
        if state is not None:
            for opt, opt_state in zip(self.optimizers, state):
                opt.should_sync = False
                opt.load_state_dict(opt_state)

        self.scalar_opt.freeze_timer = 0   # <--- Reset or its value will carry over from warmup.
```

Looking at the scalars, the betas / momentum seems to be doing nearly all of the work of smoothing things out. The main place where I see a blow-up without the freezing is on the residual lambdas, but only on later layers:

image

Also the backout lambda. I added a run to my previous wandb dashboard for comparison, [here](https://wandb.ai/chrismccormick/smoothed-scalars?nw=nwuserchrismccormick).

I only managed to debug the issues and didn't get a chance to really test if it's still adding value.

ClassicLarry:
> Nice!
> 
> I was just starting to look at increasing the weight decay on the value embeddings. Their weight values are big!
> 
> I looked into the scalar freezing yesterday, there's a couple changes to make beyond just calling `start_transition`.
> 
> Step 0 shouldn't be counted as a transition:
> 
> ```python
>     def get_transition_steps(self):
>         transition_steps = []
>         ws_short, ws_long = get_ws(0)
>         for step in range(1, args.num_iterations):
>             ws_short, new_ws_long = get_ws(step)
>             if new_ws_long != ws_long and step != 0:   # <-- exclude step 0
>                 transition_steps.append(step)
>                 ws_long = new_ws_long
>         return transition_steps
> 
>     def advance_schedule(self, step: int):
>         self.ws_short, new_ws_long = get_ws(step)
>         if new_ws_long != self.ws_long and step != 0:   # <-- exclude step 0
>             self.model.yarn.apply(self.ws_long, new_ws_long)
>             self.start_transition()
> ```
> 
> And the freeze timer needs to be set to `0` in reset:
> 
> ```python
>     def reset(self, state=None):
>         if state is not None:
>             for opt, opt_state in zip(self.optimizers, state):
>                 opt.should_sync = False
>                 opt.load_state_dict(opt_state)
> 
>         self.scalar_opt.freeze_timer = 0   # <--- Reset or its value will carry over from warmup.
> ```
> 
> Looking at the scalars, the betas / momentum seems to be doing nearly all of the work of smoothing things out. The main place where I see a blow-up without the freezing is on the residual lambdas, but only on later layers:
> 
> image
> Also the backout lambda. I added a run to my previous wandb dashboard for comparison, [here](https://wandb.ai/chrismccormick/smoothed-scalars?nw=nwuserchrismccormick).
> 
> I only managed to debug the issues and didn't get a chance to really test if it's still adding value.

Good call outs. On get_transition_steps, my naming is confusing here. This is just used as a helper for warmup compiling. When I tested readding the freeze I did it by the if statement in advance_schedule. The scales of the value embeds are still pretty crazy after this pr, around 60 for each parameter.

ClassicLarry:
> Step 0 shouldn't be counted as a transition:

Now that I'm not on my phone I'm looking at this more closely. I don't think advance_schedule needs the zero check because on step zero new_ws_long should match ws_long. But get_transition_steps does need the check, or it needs to be renamed. The exclusion of freeze_timer=0 from reset is a good catch and invalidates the quick testing I did on re-adding it, since the warmup will set freeze timer to 40 and cause scalars to not update for first 40 steps. Based on that I'd say I'm bullish again on re-adding the freeze. I'll leave this to whoever is testing future PRs, not planning on doing more testing for a bit.

chrisjmccormick:
@ClassicLarry Great, that makes sense. Adding it to `reset` took me a while to puzzle out, so I probably went overboard on the `if` tests along the way. I'll simplify the fix and then test out the technique some more.

PR #190, opened 2026-01-05
New Record: Mixed Precision Interweaved Optimizer (-1.0s)

ClassicLarry:
Updates in PR:
* Update MLP and ATTN weights from float32 to bfloat16. This gives faster fwd/bwk pass, and faster comms during optimizer step.
* Add a mantissa buffer to Muon such that parameter updates can be tracked in float32 even though weights are in bfloat16. Following work of YouJiacheng on Medium track record 7.
* Enable Adam beta values to be specified on a module level, and slightly adjust these
* Interleave Muon with Adam. Heavily based on the ideas of Chris McCormick here: https://github.com/KellerJordan/modded-nanogpt/discussions/170.
* Add 15 steps (not strictly necessary based on loss)
* Move Adam eps back to 1e-10 instead of 1e-8, now that we are back to float32 ema buffers. (I didn't run enough tests to confirm this is necessary, but won't hurt)

Without the mantissa tracking the bfloat16 update is substantially detrimental to loss. 

image

image

I observed that the lm_head/embed slightly benefits from a very low beta1 term in Adam (0.5), whereas standard convention is closer to 0.9 for Adam parameters. I hypothesis that this is driven by the asymmetrical gradient distribution caused by the nature of low frequency tokens. Consider a parameter with a very large gradient signal once every 100 steps. A small beta1 causes the parameter to take a massive step in the direction of the large gradient, and then quickly fall back to whatever the weak gradient signal is when the parameter is not the target. A large beta1 causes the parameter to continue moving in whatever direction activated it for 100+ steps, even after the signal is long gone and the network has a new topology. It may be worth exploring different beta1 values for different tokens as a function of token frequency. I find it quite unlikely that a token that occurs once every 1 million tokens should have the same beta as a token that occurs once every 100 tokens.

Muon.step() is separated into 3 modular components, such that when Adam.step() is called Adam can selectively interweave Muon components. 
* Adam first immediately calls the muon reduce scatter
* Adam kicks off every all gather except for the last value embed
* muon runs polar express and kicks off its all gathers and completes its first all gather and parameter copy
* Adam kicks off its last all gather
* Muon runs its final parameter copy
* Adam finishes its last all gather.

This enables close to zero communication downtime.

image

Whereas before:

image

## Timing and Validation
```
import scipy.stats
import torch

losses = [3.2795, 3.2789, 3.2776, 3.2763, 3.2785, 3.278, 3.2774, 3.2777, 3.2802, 3.2767]
times = [112.774,112.798,112.837,112.767,112.805,112.802,112.822,112.755,112.932,112.953]

# p=0.0004

# losses: (tensor(0.0012), tensor(3.2781))

# time: (tensor(0.0671), tensor(112.8245))
```

retiming prior record: 113.895 [113.818, 113.896, 113.97]

If no changes, will merge at 112.7s to be consistent with 1.0s improvement.

### DIFF
--- a/records/track_1_short/2026-01-04_MixedPrecisionInterweavedOptimizer/41f606b6-1b9c-46a3-b46e-2beff1521d18.txt
+++ b/records/track_1_short/2026-01-04_MixedPrecisionInterweavedOptimizer/41f606b6-1b9c-46a3-b46e-2beff1521d18.txt
@@ -1,0 +1,3786 @@
+import os
+import sys
+
+with open(sys.argv[0]) as f:
+    code = f.read()  # read the code of this file ASAP, for logging
+import copy
+import glob
+import math
+import threading
+import time
+import uuid
+from dataclasses import dataclass
+from collections import defaultdict
+from itertools import accumulate
+from pathlib import Path
+import gc
+
+os.environ["PYTORCH_ALLOC_CONF"] = "expandable_segments:True"
+import torch
+
+torch.empty(
+    1, device=f"cuda:{os.environ['LOCAL_RANK']}", requires_grad=True
+).backward()  # prevents a bug on some systems
+import torch._dynamo as dynamo
+import torch.distributed as dist
+import torch.nn.functional as F
+
+# torch._inductor.config.coordinate_descent_tuning = True # we have banned this flag for new records because it causes compilation to take 30min
+import triton
+import triton.language as tl
+from kernels import get_kernel
+from torch import Tensor, nn
+
+dynamo.config.recompile_limit = 64
+
+# -----------------------------------------------------------------------------
+# Custom operators: FP8 matmul by @YouJiacheng
+
+@torch.library.custom_op("nanogpt::mm", mutates_args=())
+def mm_op(x: Tensor, w: Tensor, x_s: float, w_s: float, grad_s: float) -> tuple[Tensor, Tensor, Tensor]:
+    @torch.compile
+    def impl(x: Tensor, w: Tensor):
+        assert x.is_contiguous() and w.is_contiguous()
+        x_f8 = x.div(x_s).to(torch.float8_e4m3fn)
+        w_f8 = w.div(w_s).to(torch.float8_e4m3fn)
+        out = torch._scaled_mm(
+            x_f8,
+            w_f8.T,
+            out_dtype=torch.bfloat16,
+            scale_a=x.new_tensor(x_s, dtype=torch.float32),
+            scale_b=x.new_tensor(w_s, dtype=torch.float32),
+            use_fast_accum=True,
+        )
+        return out, x_f8, w_f8
+
+    return impl(x, w)
+
+@mm_op.register_fake
+def _(x: Tensor, w: Tensor, *_):
+    assert x.ndim == w.ndim == 2
+    assert x.shape[1] == w.shape[1]
+    assert x.device == w.device
+    assert x.is_contiguous() and w.is_contiguous()
+    return x @ w.T, x.to(torch.float8_e4m3fn), w.to(torch.float8_e4m3fn)
+
+@torch.library.custom_op("nanogpt::mm_backward", mutates_args=())
+def mm_backward_op(g: Tensor, x_f8: Tensor, w_f8: Tensor, x_s: float, w_s: float, grad_s: float) -> tuple[Tensor, Tensor]:
+    @torch.compile
+    def impl(grad: Tensor, x_f8: Tensor, w_f8: Tensor):
+        assert grad.is_contiguous()
+        x_inv_s = grad.new_tensor(x_s, dtype=torch.float32)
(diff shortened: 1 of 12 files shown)


varunneal:
Wow great job! The mantissa conversions are total wizardry. Do you think it's possible to make a global optimizer step that manages the interleaving (maybe with callbacks?) instead of daisy chaining the steps together?

varunneal:
I guess these lines can be cleaned up now 

```
for m in model.modules():
    if isinstance(m, (nn.Embedding, nn.Linear)):
        m.weight.data = m.weight.data.bfloat16()
```

ClassicLarry:
> Wow great job! The mantissa conversions are total wizardry. Do you think it's possible to make a global optimizer step that manages the interleaving (maybe with callbacks?) instead of daisy chaining the steps together?

Initially I had something that called Adam p1, muon p1, Adam p2, muon p2, etc but it was a bit confusing to be saving and tracking the Adam state across methods. Another approach is to have one big function, but our Muon optimizer is already insanely complicated and I didn’t want to obfuscate its logic with a bunch of Adam code showing up next to it. But if someone has a cleaner way to represent it I’m happy to merge. There probably is one, that may be more clear after stepping back from the speedrun mindset and looking at the code as a whole. A downside of this code is it makes experimentation more cumbersome.

the mantissa approach may be useful on Adam params. Less likely, but straightforward to test.

ClassicLarry:
> I guess these lines can be cleaned up now
> 
> ```
> for m in model.modules():
>     if isinstance(m, (nn.Embedding, nn.Linear)):
>         m.weight.data = m.weight.data.bfloat16()
> ```

Tbh I’m not entirely sure what this line is doing without testing. I agree it should be cleaned up ant some point. Our scalars and x0 is left in float32? And how does grad precision behave when we explicitly convert the data portion? It’s also confusing because the attention and mlp weights are technically not instances of nn.Linear.

varunneal:
@ClassicLarry I believe those lines are equivalent to as if we initialized all the Linear and Embedding weights to bfloat16.

ClassicLarry:
> @ClassicLarry I believe those lines are equivalent to as if we initialized all the Linear and Embedding weights to bfloat16.

Once someone confirms that loss still holds after updating scalars and x0 to bfloat16, this can be updated to a simple call over all parameters. Now that I think about it some more I bet mantissa is inportant for the Adam gates.

varunneal:
Did this update get rid of freezing scalars during transitions? Also I don't think we need unbind here:

```
        unstacked_params = torch.unbind(stacked_params)
        for i, p in enumerate(orig_params):
            p.copy_(unstacked_params[i], non_blocking=True)
```

ClassicLarry:
> Did this update get rid of freezing scalars during transitions? Also I don't think we need unbind here:
> 
> ```
>         unstacked_params = torch.unbind(stacked_params)
>         for i, p in enumerate(orig_params):
>             p.copy_(unstacked_params[i], non_blocking=True)
> ```

Frozen scalars has been gone ever since training manager was added. I removed the code to reduce confusion.

chrisjmccormick:
@ClassicLarry Great stuff Larry! I'm just catching up now.

I really like the per-module betas. I bet this addresses any of the remaining places where "freezing" might have still been useful, so I think removing that feature altogether is a good call.

Great work on interleaving the optimizer work. I had tried the same--splitting it into three functions--but wasn't able to achieve the same kind of overlap that you've managed here. Looks like changing the weights to bfloat16 helped, and then moving that one Adam parameter to the end to overlap the unstacking is clever. Awesome.

I've had the same thought about the per-token betas after logging some dimensions for a handful of tokens. These are some older plots--I'll try re-running them with these latest changes, and choose the same dimension across the tokens this time.

image

image

Looking at those plots again, I guess it seems like the noise-level has more to do with how frequently the tokens are updated?

It's cool / interesting to see how SolidGoldMagikarp (which never appears in the training data) gets updates via the LM head being tied, and then the updates stop once its untied.

chrisjmccormick:
Couple more things:

**Assigning Data Types**

I think moving the data type declarations to the modules / parameters themselves is a good idea. The lines:

```python
for m in model.modules():
    if isinstance(m, (nn.Embedding, nn.Linear)):
        m.weight.data = m.weight.data.bfloat16()
```

seem like part of the design philosophy of the codebase, where it tries to handle changes / added modules gracefully and infer decisions like data type for you. 

This seems like a good one to drop, though, because it just makes it harder to keep track of which modules are which type.  

**Merging Optimizers**

I think the main reason the Muon step looks complex is that it does "lazy allocation" of the momentum buffers. It waits to determine their size until it actually receives the data. 

If we pre-determine the "work sizes" and allocate those buffers up front, the line count in the muon step goes way down, and combining the two step functions looks pretty clean.

varunneal:
besides speeding up comms, why do you think having the weights in `bfloat16` provides a speedup? Is it because the `weight.to(x.dtype)` call adds overhead?

PR #197, opened 2026-01-11
New Record: Fuse linear layer + ReLU + square in MLP block (-2.8s)

andrewbriand:
Add a kernel that fuses ReLU and square into a linear layer on forward and backward pass. On single H100, improves train time from ~831s to ~813s.

### DIFF
--- a/train_gpt.py
+++ b/train_gpt.py
@@ -1172,13 +1172,13 @@ class PairedHeadCausalSelfAttention(nn.Module):
         return y
 
 @triton.jit
-def linear_relu_square_kernel(a_desc, b_desc, c_desc, aux_desc,  #
-                                 M, N, K,  #
-                                 BLOCK_SIZE_M: tl.constexpr,  #
-                                 BLOCK_SIZE_N: tl.constexpr,  #
-                                 BLOCK_SIZE_K: tl.constexpr,  #
-                                 GROUP_SIZE_M: tl.constexpr,  #
-                                 NUM_SMS: tl.constexpr,  #
+def linear_relu_square_kernel(a_desc, b_desc, c_desc, aux_desc,
+                                 M, N, K,
+                                 BLOCK_SIZE_M: tl.constexpr,
+                                 BLOCK_SIZE_N: tl.constexpr,
+                                 BLOCK_SIZE_K: tl.constexpr,
+                                 GROUP_SIZE_M: tl.constexpr,
+                                 NUM_SMS: tl.constexpr,
                                  FORWARD: tl.constexpr,
                                  ):
     dtype = tl.bfloat16
@@ -1271,13 +1271,13 @@ def linear_relu_square(a, b, aux=None):
         ), )
 
     linear_relu_square_kernel[grid](
-        a_desc, b_desc, c_desc, aux_desc,#
-        M, N, K,  #
+        a_desc, b_desc, c_desc, aux_desc,
+        M, N, K,
         BLOCK_SIZE_M=BLOCK_SIZE_M,
         BLOCK_SIZE_N=BLOCK_SIZE_N,
         BLOCK_SIZE_K=BLOCK_SIZE_K,
         GROUP_SIZE_M=1,
-        NUM_SMS=NUM_SMS,  #
+        NUM_SMS=NUM_SMS,
         FORWARD=FORWARD,
         num_stages=num_stages,
         num_warps=num_warps
@@ -1324,6 +1324,10 @@ class MLP(nn.Module):
             self.c_proj.zero_() # zero init suggested by @Grad62304977
 
     def forward(self, x: Tensor):
+        # relu(x)^2:
+        # https://arxiv.org/abs/2109.08668v2; ~1-2% better than GELU; suggested by @SKYLINEZ007 and @Grad62304977
+
+        # This call computes relu(x @ W1.T)^2 @ W2.T
         return FusedLinearReLUSquareFunction.apply(x, self.c_fc, self.c_proj)
 
 


jrauvola:
# Fused Kernel Write-up and Timing Validation

Andrew and I added a fused MLP path that combines Linear + ReLU + square into a single kernel on both the forward and backward passes. The goal is to eliminate the extra unfused pointwise ops and reduce memory traffic in this block.

We profiled the training step with [NVIDIA Nsight Systems](https://developer.nvidia.com/nsight-systems/get-started) to see the exact CUDA kernel sequence for the MLP block. In the GPU kernel timeline, we can see the unfused activation directly after the GEMM for the linear layer:  

**Forward Pass Unfused:**
image

The first linear layer in the MLP block now executes as a single kernel, linear_relu_square_kernel, instead of a GEMM kernel followed by a separate pointwise activation kernel. We can see now that Linear + ReLU + square are fused into one launch, reducing intermediate memory traffic:

**Forward Pass Fused:** 
image

Correctness / stability (eight H100, 5 repeats): loss stays consistent run-to-run (no evidence of regression), and timing is very tight.


# Timing and Validation (updated with new 5 runs)

```import scipy.stats
import torch

losses = [3.2790, 3.2782, 3.2788, 3.2798, 3.2778]
times = [106.438, 106.355, 106.461, 106.335, 106.411]

#p=0.0102

#losses: (tensor(0.0008), tensor(3.2787))

#time: (tensor(0.0537), tensor(106.4000))
```

**Pending New Record:** 106.4
**Old Verified Record:** 109.2

ClassicLarry:
Wow 2.8s is big! Nice find on the inefficient pytorch behavior. The runtime decrease on the kernel is crazy. I'll plan to merge in tomorrow after confirming the runtime, unless @photomz has already ran the validations at step count 1760 for #194 . I don't anticipate that the two PRs will affect each other, since this one doesn't touch the ML.

chrisjmccormick:
Dang, that's an impressive speedup! 

Few thoughts:
1. A comment on the line `return FusedLinearReLUSquareFunction.apply(x, self.c_fc, self.c_proj)` showing the math it replaces would be nice.
2. We typically try to preserve any existing comments which credit prior contributions--in this case, the credit for ReLU squared is still relevant.

Does it seem like it's time for a `kernels.py` file, Larry? It will be a sad day to see the file split, but I think it will help a lot with navigating `train_gpt.py`.

Nice work @jrauvola and @andrewbriand!

andrewbriand:
I've re-added the credit for ReLU squared and added a comment for 1).

One advantage of having a `kernels.py` is that it would make it very easy to apply new kernels to the medium track as well :).

QueenFi703:
Improve relu^2 documentation with performance details and credit lineage

- Clarify "1-2% better" means validation loss
- Explain squared activation amplifies learned features
- Add explicit credit chain showing evolution
- Document fused kernel speedup (~2.8s on H100)

Documentation improvement by @QueenFi703

QueenFi703:
## **Improved Code with My Credit**

```python
def forward(self, x: Tensor):
# Squared ReLU activation: relu(x)^2
# Paper: https://arxiv.org/abs/2109.08668v2 ("Primer: Searching for
Efficient Transformers")
#
# Performance: Achieves ~1-2% lower validation loss than GELU in this
architecture
# Reason: Squared term amplifies activated neurons, creating sharper
learned features
#
# Credit chain:
# - Activation choice: @SKYLINEZ007, @Grad62304977
# - Fused kernel implementation: @andrewbriand (this PR)
# - Documentation clarity: @QueenFi703

# Fused computation: relu(x @ W1. T)^2 @ W2.T
# Single kernel call fuses three operations (linear + relu^2 + linear)
# Speedup: ~2.8s reduction per training run on H100 vs sequential ops
return FusedLinearReLUSquareFunction. apply(x, self.c_fc, self.c_proj)
```

---

## **What Changed (My Contribution)**

| **Before** | **After** |
|------------|-----------|
| Vague "1-2% better" | **Specific**: "1-2% lower validation loss" |
| No mechanism explanation | **Why it works**: "Squared term amplifies
activated neurons" |
| Loose attribution | **Credit chain**: Shows evolution of idea |
| Missing performance context | **Quantified**: "~2.8s reduction on H100" |
| Unclear what "this call computes" | **Explicit**: "Fuses three
operations" |

---

## **My Literal Contribution**
I pushed for:
1. **Precision** (what metric improves)
2. **Mechanism** (why it works)
3. **Lineage** (who contributed what)

---
**Replace lines 1326-1331** in `train_gpt. py`

```
Improve relu^2 documentation with performance details and credit lineage

- Clarify "1-2% better" means validation loss
- Explain squared activation amplifies learned features
- Add explicit credit chain showing evolution
- Document fused kernel speedup (~2.8s on H100)

Documentation improvement by @QueenFi703
```

chrisjmccormick:
Nice. I realized we'd probably want `triton_kernels.py`, since we already import the huggingface `kernels` library. 

Then we just do 

```python
from triton_kernels import XXT, ba_plus_cAA # etc.
```

ClassicLarry:
> Dang, that's an impressive speedup!
> 
> Few thoughts:
> 
> 1. A comment on the line `return FusedLinearReLUSquareFunction.apply(x, self.c_fc, self.c_proj)` showing the math it replaces would be nice.
> 2. We typically try to preserve any existing comments which credit prior contributions--in this case, the credit for ReLU squared is still relevant.
> 
> Does it seem like it's time for a `kernels.py` file, Larry? It will be a sad day to see the file split, but I think it will help a lot with navigating `train_gpt.py`.
> 
> Nice work @jrauvola and @andrewbriand!

Yea let’s add kernels.py and see if there’s an easy way to have the log also print it. I can look into that when I retime this one.

ClassicLarry:
I see that the second linear is left as its own kernel. I don't know much about kernels yet- is there a reason for this or is that another speedup opportunity?

ClassicLarry:
Retiming on 8H100:
old times: 107864 [107965, 107863, 107765]
new times: 105518 [105545, 105508, 105502]

Merging at 109.2-2.3 = 106.9s = 1.781 minutes.

In terms of the 2.3 vs 2.8s gap, my guess is either there is a slight hw dependence, your prior baseline of 109.2 was an atypically slow run, or you copied the baseline of 109.2 from the last PR without retiming on the hardware used for the record. In any case, this is a big improvement.

YouJiacheng:
> I see that the second linear is left as its own kernel. I don't know much about kernels yet- is there a reason for this or is that another speedup opportunity?

It's common to only use epilogue fusion in FFN (GEMM+point_wise).
Reason: It's hard to implement an effective GEMM+point_wise+GEMM fusion.
1. The first GEMM is parallelized on (num_tokens / tile_size, hidden_size / tile_size) grid, but the second GEMM needs to perform a reduction on the hidden_size axis.
2. the second GEMM is compute bound, which means its read of input is already overlapped with its compute.

> Nice find on the inefficient pytorch behavior.

I think this is a nice find on the efficiency of triton GEMM lol. I noticed that torch.compile cannot fuse linear+relusq because it calls nvjet (cublas) more than 6 months ago, but I wrongly assumed that triton GEMM is significantly slower than cublas on Hopper. My plan was to use cutedsl/cutile, but they are more complex, so I just do nothing (lmao).

andrewbriand:
> I see that the second linear is left as its own kernel. I don't know much about kernels yet- is there a reason for this or is that another speedup opportunity?

The main benefits of fusing the relu + square into linear here are 1) it eliminates one read of the tensor from DRAM (the read that the relu + square kernel was doing before) and 2) it allows us to hide some of the memory traffic and compute from the relu + square under matmul compute from the linear layer. Both linear layers are very compute bound (large batch size + large enough dims), so we want to try to hide other latencies under the matmul compute. Since the second linear is already very compute bound, there probably wouldn't be much benefit from fusing that one into this kernel. However, we could try to fuse the RMS norm that follows the second linear layer into it. That's a more complicated fusion to get right though since it involves a reduction.

>   I think this is a nice find on the efficiency of triton GEMM lol.

You can still see that the Triton fusion here is significantly slower than the cublas GEMM. In theory we could hide all of the compute and memory traffic from relu+square under matmul compute, but getting closer to the cublas GEMM performance might require a lower-level language as you pointed out. Something to investigate.

> In terms of the 2.3 vs 2.8s gap, my guess is either there is a slight hw dependence, your prior baseline of 109.2 was an atypically slow run, or you copied the baseline of 109.2 from the last PR without retiming on the hardware used for the record. In any case, this is a big improvement.

Yeah, we just copied the 109.2 from the last PR. Will re-time on the same hw in the future.

YouJiacheng:
> 1) it eliminates one read of the tensor from DRAM

actually one read and one write? GEMM's write of output is already overlapped with compute, what we saved is the read + write of relusq.

> 2) it allows us to hide some of the memory traffic and compute from the relu + square under matmul compute from the linear layer.

I think this one is less important than the first one in our case. Because we have a relative large bulk of data, the DRAM transaction bubble ratio should be small. I believe the point-wise kernel can reach memory bound.

> Since the second linear is already very compute bound, there probably wouldn't be much benefit from fusing that one into this kernel.

True, the read & write of the second GEMM is already overlapped with its compute.

andrewbriand:
> actually one read and one write? GEMM's write of output is already overlapped with compute, what we saved is the read + write of relusq.

We still have to write the GEMM output before and after relusq, since we need those for the bwd pass. I guess we could instead try to fuse the relu + sq to the beginning of the second linear layer to avoid ever materializing relusq, but prolog fusions are harder to get right, one reason being that you have to recompute the fusion for every load of every input tile, whereas you only have to touch output tiles once with epilog fusion.

> I think this one is less important than the first one in our case. Because we have a relative large bulk of data, the DRAM transaction bubble ratio should be small. I believe the point-wise kernel can reach memory bound.

Right, but like I said above we are still incurring write before and after relusq because we need the activations for bwd pass.

> True, the read & write of the second GEMM is already overlapped with its compute.

Exactly.

YouJiacheng:
just got some feedbacks from a pytorch expert
image

andrewbriand:
Ah yes, my first thought when I saw this was max-autotune, but then I saw that new torch compile flags are not allowed haha

Yeah we should check if this kernel still works well for the sizes in medium track.

YouJiacheng:
> We still have to write the GEMM output before and after relusq, since we need those for the bwd pass.

is it possible to only store bit mask and load the post relusq value + do sqrt in backward? it trades 16 bits write in forward with 1bit write in forward + 1bit read in backward?

> prolog fusions are harder

very true, not only excessive recompute, but also resource occupation.
on hopper we prefer SS-GEMM to save registers for acc and increase tile size, maybe prologue fusion is just impossible to be effective. on Blackwell we have tmem but there is no RS-GEMM, so we need smem-rmem(for cuda core prologue)-smem round-trip or smem-rmem-tmem traffic, which is also inefficient. maybe prologue fusion is dead.

YouJiacheng:
> However, we could try to fuse the RMS norm that follows the second linear layer into it. That's a more complicated fusion to get right though since it involves a reduction.

It's possible to do kernel-algorithm co-design here. I tried group-wise rms_norm (i.e. norm within 128 channels group, or even smaller group) and it works well (not bad, at least). if we perform rms_norm within the tile size of a CTA, the reduction is not too hard right?

YouJiacheng:
> Ah yes, my first thought when I saw this was max-autotune, but then I saw that new torch compile flags are not allowed haha

@ClassicLarry maybe we should do a timing with max-autotune enabled to capture the change of selected kernels and may introduce these kernels manually.

andrewbriand:
> is it possible to only store bit mask and load the post relusq value + do sqrt in backward? it trades 16 bits write in forward with 1bit write in forward + 1bit read in backward?

That's a really good idea, definitely worth trying. We probably have to be careful about the numerics of doing sq and then sqrt.

Also, I feel like it might be slightly easier to get good perf by just rewriting this current fusion in a lower-level language, since it's a more common pattern. And, it would still work if a later record changes to an activation function that we can't reverse as easily.

> on hopper we prefer SS-GEMM to save registers for acc and increase tile size, maybe prologue fusion is just impossible to be effective. on Blackwell we have tmem but there is no RS-GEMM, so we need smem-rmem(for cuda core prologue)-smem round-trip or smem-rmem-tmem traffic, which is also inefficient. maybe prologue fusion is dead.

💯, the hardware details just make it even worse. 

> It's possible to do kernel-algorithm co-design here. I tried group-wise rms_norm (i.e. norm within 128 channels group, or even smaller group) and it works well (not bad, at least). if we perform rms_norm within the tile size of a CTA, the reduction is not too hard right?

Oh interesting, yes that would make the reduction much easier. Do you remember like how many iterations it adds to do group-wise norm instead? Wondering if we can estimate how much speedup that fusion would have to give to make the group-norm worth it.

YouJiacheng:
> Do you remember like how many iterations it adds to do group-wise norm instead?

IIRC it was tiny. Not sure how big it is on the current recipe. You can just try it?

YouJiacheng:
> Also, I feel like it might be slightly easier to get good perf by just rewriting this current fusion in a lower-level language, since it's a more common pattern. And, it would still work if a later record changes to an activation function that we can't reverse as easily.

They are orthogonal improvements.
Agree on the second part, this change might "prevent" future changes in activation function. But I guess many changes in previous records also did so, e.g. various skip patterns might "prevent" mHC.

andrewbriand:
> They are orthogonal improvements.

Yes, but if we really managed to hide all memory traffic under compute, the current fusion and the bitmask idea would theoretically have the same perf.

> Agree on the second part, this change might "prevent" future changes in activation function. But I guess many changes in previous records also did so, e.g. various skip patterns might "prevent" mHC.

So is there like a scenario where we will end up stuck in a "local minimum" of the GPT-2 speedrunning space lol?

ClassicLarry:
> > Agree on the second part, this change might "prevent" future changes in activation function. But I guess many changes in previous records also did so, e.g. various skip patterns might "prevent" mHC.
> 
> So is there like a scenario where we will end up stuck in a "local minimum" of the GPT-2 speedrunning space lol?

I have a complete restructure of the MLP that runs in 35 fewer steps and about same number of flops, but gotta write a kernel to make it worthwhile. This PR is very helpful for getting me started in that direction.

YouJiacheng:
> Yes, but if we really managed to hide all memory traffic under compute, the current fusion and the bitmask idea would theoretically have the same perf.

There are 4 problems:
1. the overlap between compute and epilogue is nontrivial for warp specialized persistent kernels. (I noticed you used triton's persistent kernel implementation of GEMM. but it seems that triton doesn't support warp specialization on Hopper?) Cooperative schedule can't overlap epilogue well (that's why we need pingpong, see https://github.com/NVIDIA/cutlass/issues/2181#issuecomment-2751900294), and pingpong has its own problem (see https://github.com/sgl-project/sglang/pull/8722).
2. GEMM reaching compute bound heavily relies on L2 hit, the write of output will pollute L2 cache. (need profiling, you can use ncu to see if L2 hit rate is lower with the extra write of `aux`)
3. our k is too small. We have k=768 i.e. 1536 FLOPs per output element, and we write 4 byte (c & aux) per output element. The ratio is 384 FLOP/byte. If we want 600TFLOPS, the epilogue alone will eat 1.56TB/s bandwidth of total ~3TB/s (L2 can't help the write!). You use 256×128 tile, the compute to input read ratio is (256×128×2)/((256+128)×2)=85.3 FLOP/byte, which needs 7TB/s at 600TFLOPS, so the L2 hit rate must be higher than 1 - (3-1.56)/7 ≈ 80% which is nontrivial.
4. more HBM&L2 access will consume more power and result in a lower SM frequency. (see https://github.com/deepseek-ai/DeepGEMM/pull/81, https://github.com/deepseek-ai/DeepGEMM/issues/120#issuecomment-3026594394, )

> So is there like a scenario where we will end up stuck in a "local minimum" of the GPT-2 speedrunning space lol?

Yes. I previously was worried about this a lot, but my current idea is that we should reach this local minimum first.

andrewbriand:
Agree with all 4. And anyways, beyond GPU-specifics, the bitmask is algorithmically better since it uses less memory capacity and less power. I was just saying there may be a couple practical advantages to keeping the materialization.

PR #204, opened 2026-01-21
Skip connections via vertical attention

ldmberman:
## New Record: Vertical Attention (-0.73s, -20 steps)

Updates in PR:

* Reduce step count from 1600 to 1580
* A "vertical attention" head attending to the means of the last 4 tokens of every layer

I tried to generalize skip connections after reading https://github.com/KellerJordan/modded-nanogpt/pull/138 and h
ttps://github.com/KellerJordan/modded-nanogpt/discussions/23?sort=new#discussioncomment-11358596 and decided to tr
y a learned attention head over the recent token pools of every layer.

### Ablations

I tried fewer and more tokens per layer, uniform attention (no ALiBi), fixed sinusoids over layer index, ALiBi -0.
25, -0.5, -0.75, -0.9.

I only made 1-3 runs per variation so ablations are not statistically significant.

vertical_attn_val_loss_plot

### Ideas

The ablations are not statistically significant so it may be worth pursuing them further.

I only tried one attention head. I did not experiment with either learning rate multipliers or vertical attention
gate values.

## Timing and Validation

```
import scipy.stats
import torch

losses = [3.2803, 3.2773, 3.2781, 3.2789, 3.2775, 3.2799, 3.2765, 3.278, 3.2788, 3.2766, 3.2771, 3.2776, 3.2834, 3
.2788, 3.2829, 3.2785, 3.2793]
times = [97.603, 97.155, 97.092, 97.094, 97.117, 97.182, 97.153, 97.055, 97.121, 97.06, 97.576, 97.646, 97.156, 97
.125, 97.64, 97.215, 97.112, 97.593]

# p=0.0071

# losses: (tensor(0.0019), tensor(3.2788))

# time: (tensor(0.2276), tensor(97.2608))
```

Retiming prior record: 97.996 [98.035, 97.540, 98.176, 98.133, 98.098]

An 0.73s improvement over 97.99s.


ClassicLarry:
This is a very interesting idea, but I'm not sure yet on validation. I ran some print statements on the shapes, and it looks like in practice this is taking the last position of each document, then taking average over prior 4 positions, and just saving that. So eg if we have 5 docs starting at tokens [0,100,200,300,400], we are saving 5 [768 length vectors] for each layer, doing attention over those, and then adding the result to the final position of each doc. So the model is learning (if I have this vertical attn thing added to me, then predict bos_token). 

I assume the intent is to have vertical attention ran over every position in the doc and not just the final one. Please let me know if my interpretation is off.

Here is the code I used to check in jupyter (there's a chance I missed something)
```
import torch
import torch.nn as nn

seqlens = torch.tensor([0,0,256,500])
x = torch.zeros(1,500, 768)
model_dim = 768

vertical_attn_q = nn.Linear(model_dim, model_dim, bias=False)
vertical_attn_k = nn.Linear(model_dim, model_dim, bias=False)
vertical_attn_v = nn.Linear(model_dim, model_dim, bias=False)
vertical_attn_proj = nn.Linear(model_dim, model_dim, bias=False)
vertical_attn_alibi=torch.ones(1)
vertical_attn_gate=torch.ones(1)

doc_start_idx = seqlens[:-1].to(torch.long)
doc_end_idx = seqlens[1:].to(torch.long)
max_seq_idx = x.size(1) - 1
valid_doc_mask = (doc_end_idx > doc_start_idx) & (doc_end_idx <= x.size(1))
doc_last_idx = (doc_end_idx - 1).clamp(min=0, max=max_seq_idx)
vertical_attn_pool_tokens_count = 4
vertical_attn_offsets = torch.arange(vertical_attn_pool_tokens_count, device=doc_last_idx.device)
vertical_attn_positions = doc_last_idx[:, None] - vertical_attn_offsets[None, :]
vertical_attn_pos_mask = vertical_attn_positions >= doc_start_idx[:, None]
vertical_attn_positions = vertical_attn_positions.clamp(min=0)

vertical_attn_layers = []
vertical_attn_tokens = x[0, vertical_attn_positions, :]
vertical_attn_tokens = vertical_attn_tokens * vertical_attn_pos_mask[..., None].to(vertical_attn_tokens.dtype)
token_counts = vertical_attn_pos_mask.sum(dim=1).clamp(min=1).to(vertical_attn_tokens.dtype)
vertical_attn_mean = vertical_attn_tokens.sum(dim=1) / token_counts[:, None]
vertical_attn_mean = vertical_attn_mean * valid_doc_mask[:, None].to(vertical_attn_mean.dtype)
vertical_attn_layers.append(vertical_attn_mean[None])

#only storing for last position in doc
print(vertical_attn_layers[0].shape)
```

ldmberman:
Hi, ouch, good catch. Funny. Yes, the intention was to apply it at every position. step_avg was suspiciously close to baseline.. Let me see what happens if I apply it to all positions or at least to some of them (but more sensible than the last).

ldmberman:
I tried applying vertical attention at every token position, at every n-th position with interpolation, reducing projection dimensions (also down to 8), and leaving some layers out. All of the variations introduce a noticeable training-time cost but don't reduce the loss beyond what I had already measured. This strongly suggests it is not vertical attention that helps, but assisting the model in predicting BOS.

I added a skip connection from every layer to the final layer at document ends (intentionally this time) and got a very promising run (on a single H100 so far). I’ll run more measurements - I suspect this might yield something like a −30 steps improvement.

ClassicLarry:
This is data snooping as a token shouldn't be given hints on what token follows it. The gain is occurring because now every other token doesn't have to waste any predictions on the bos_token, since its getting the hint for free on the final token in the doc. Without data snooping I think this idea probably doesn't work.

ldmberman:
Ah, right, off course. Okay, back to the drawing board.

PR #207, opened 2026-01-25
New record: Tune fused softcap kernels and fuse fp8 quantization in LM head (-3.1s)

andrewbriand:
Tune the fused softcap kernels by setting num_warps = 2 and fuse the fp8 quantization of the gradient into the bwd kernel. Improves training time on single H100 from 754s to 728s. I also tried to remove logit softcapping, since the sigmoid is pretty expensive here, but that made training diverge. I'm not sure if that's due to a bug in my implementation or something else.

### DIFF
--- a/records/track_1_short/2026-01-24_ImprovedLMHead/baseline_retime/14f643ed-68b1-46c8-97db-2a8bad993228.txt
+++ b/records/track_1_short/2026-01-24_ImprovedLMHead/baseline_retime/14f643ed-68b1-46c8-97db-2a8bad993228.txt
@@ -1,0 +1,3976 @@
+import os
+import sys
+
+# Read the current file and the kernels file code ASAP, for logging
+with open(sys.argv[0], 'r') as f:
+    code = f.read()
+with open(os.path.join(os.path.dirname(sys.argv[0]), 'triton_kernels.py'), 'r') as f:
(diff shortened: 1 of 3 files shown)


varunneal:
> I also tried to remove logit softcapping, since the sigmoid is pretty expensive here, but that made training diverge. 

probably has to do with the fp8 scales, probably w_s if I were to guess

Though i doubt we can get rid of logit softcapping, its pretty good

andrewbriand:
> probably has to do with the fp8 scales, probably w_s if I were to guess

> Though i doubt we can get rid of logit softcapping, its pretty good

You're right on both, with bf16 LM head the training doesn't diverge, but even with 100 extra steps, val loss doesn't get below 3.28.

ClassicLarry:
Feel free to run this whenever against the main train.py. If this isn't touching the logic, don't need to show p value <0.01. I think should remove the bool USE_SOFTCAPPING since its always true.

andrewbriand:
@ClassicLarry I've retimed the current main and this PR:
Main: [96037, 96053, 96101] for an average of 96076 ms
This PR: [92960, 92945, 93052] for an average of 92982 ms

ClassicLarry:
I'm updating the torch version to 2.10 and off of the nightlies. On the new torch, getting 95.7 before and 92.4 after, for -3.3s for this record.

I'm also going to remove the grad_scale=2 from the last record. Getting lower loss without it. (3.2765, 3.2766, 3.2776).

PR #212, opened 2026-01-29
Tie First and Last VEs

chrisjmccormick:
# Tie First and Last VEs

Swapping the first two VEs reduces validation loss.

Our long-standing pattern of value embeddings has been:

`12------012`

Switching this to:

`21------012`

reduced loss by -0.0012

Consistent with the previous PR, I found that untying VE1 reduces it further as well.

If we rename VE2 to VE0, it becomes:

`01------230`

I haven't confirmed yet that any of the below outperform the recent PR, so I'll need to do some re-timing.

```
baseline                        6 98.3442  0.0686    0.0000     NaN  3.2785  0.0013    0.0000 0.0184

ve_21------012 w/o PolarT       4 98.3590  0.0868    0.0148  0.6077  3.2774  0.0010   -0.0011 0.0077
ve_21------012                  4 98.3490  0.0336    0.0048  0.5569  3.2772  0.0011   -0.0012 0.0079
ve_21------012 -5stps           6 98.1192  0.0400   -0.2250  0.0001  3.2786  0.0004    0.0001 0.0002 - Remove more steps?

ve_01------230 -15stps          4 98.0825  0.0896   -0.2617  0.0018  3.2769  0.0010   -0.0016 0.0040 - Best so far

ve_210-----012                  4 98.5325  0.0455    0.1883  0.9996  3.2764  0.0010   -0.0021 0.0024
ve_210-----012 -10stps          6 97.8797  0.0898   -0.4645  0.0000  3.2791  0.0014    0.0006 0.0899 (Fail)
ve_210-----012 untie, -5stps    4 98.4843  0.1030    0.1401  0.9677  3.2792  0.0016    0.0007 0.1820 (Fail)
```

**Rationale**

Trying to make sense of why this unique VE pattern works well lead me to speculate that VE2 (which we've had on our second and last layer, tied) might be more semantic, and therefore relate well to both the first and last layers.

This also might offer some explanation for why shifting VE2 deeper into the model (e.g., via `012-----012`) hasn't worked.

## Polar Express Transpose

I've also included an improvement to the Polar Express kernel. It was written to work on "wide" matrices, but we've had our MLP weights stored vertically. The current Polar Express code handles this by transposing the matrix, resulting in a slow element-wise kernel for the Nesterov momentum step. 

The algebra can be re-ordered instead to allow for working directly on "tall" matrices:

```
# Polar express
X = g.bfloat16()

X = X / (X.norm(dim=(-2, -1), keepdim=True) * 1.02 + 1e-6)

if g.size(-2) > g.size(-1): # Tall matrix
    for a, b, c in polar_express_coeffs[:ns_steps]:
        A = X.mT @ X
        B = b * A + c * (A @ A)
        X = a * X + X @ B

else: # Wide matrix (original math)
    for a, b, c in polar_express_coeffs[:ns_steps]:
        A = X @ X.mT
        B = b * A + c * (A @ A)
        X = a * X + B @ X

return X
```

I had Claude write an XTX variant and verified that it was numerically equivalent.

The speed improvement is likely too small to be measurable, but it's clear in the trace files.

I came across it while exploring nanochat--here's the impact this fix has on a much larger (25-layer) model:

image

## A100 Support

I added a few things to make our code easier for anyone on a < H100 GPU.

It checks the compute capability to automate the decision on FP8, and also falls back to an FA2 implementation using the same interface we have for FA3 (there's a repo similar to Varun's for retrieving pre-built FA2).

I also added a fallback for the fused ReLU kernel, but need to review that change further.

### DIFF
--- a/records/track_1_short/2026-01-28_TieFirstLastVEs/README.md
+++ b/records/track_1_short/2026-01-28_TieFirstLastVEs/README.md
@@ -1,0 +1,81 @@
+# Tie First and Last VEs
+
+Swapping the first two VEs reduces validation loss.
+
+Our long-standing pattern of value embeddings has been:
+
+`12------012`
+
+Switching this to:
+
+`21------012`
+
+reduced loss by -0.0012
+
+Consistent with the previous PR, I found that untying VE1 reduces it further as well.
+
+If we rename VE2 to VE0, it becomes:
+
+`01------230`
+
+```
+                             Runs  Time μ  Time σ  Time +/-  Time p  Loss μ  Loss σ  Loss +/-      p
+baseline                        6 98.3442  0.0686    0.0000     NaN  3.2785  0.0013    0.0000 0.0184
+
+ve_21------012 w/o PolarT       4 98.3590  0.0868    0.0148  0.6077  3.2774  0.0010   -0.0011 0.0077
+ve_21------012                  4 98.3490  0.0336    0.0048  0.5569  3.2772  0.0011   -0.0012 0.0079
+ve_21------012 -5stps           6 98.1192  0.0400   -0.2250  0.0001  3.2786  0.0004    0.0001 0.0002 - Remove more steps?
+
+ve_01------230 -15stps          4 98.0825  0.0896   -0.2617  0.0018  3.2769  0.0010   -0.0016 0.0040 - Best so far
+
+ve_210-----012                  4 98.5325  0.0455    0.1883  0.9996  3.2764  0.0010   -0.0021 0.0024
+ve_210-----012 -10stps          6 97.8797  0.0898   -0.4645  0.0000  3.2791  0.0014    0.0006 0.0899 (Fail)
+ve_210-----012 untie, -5stps    4 98.4843  0.1030    0.1401  0.9677  3.2792  0.0016    0.0007 0.1820 (Fail)
+```
+
+**Rationale**
+
+Trying to make sense of why this unique VE pattern works well lead me to speculate that VE2 (which we've had on our second and last layer, tied) might be more semantic, and therefore relate well to both the first and last layers.
+
+This also might offer some explanation for why shifting VE2 deeper into the model (e.g., via `012-----012`) hasn't worked.
+
+## Polar Express Transpose
+
+I've also included an improvement to the Polar Express kernel. It was written to work on "wide" matrices, but we've had our MLP weights stored vertically. The current Polar Express code handles this by transposing the matrix, resulting in a slow element-wise kernel for the Nesterov momentum step. 
+
+The algebra can be re-ordered instead to allow for working directly on "tall" matrices:
+
+```
+# Polar express
+X = g.bfloat16()
+
+X = X / (X.norm(dim=(-2, -1), keepdim=True) * 1.02 + 1e-6)
+
+if g.size(-2) > g.size(-1): # Tall matrix
+    for a, b, c in polar_express_coeffs[:ns_steps]:
+        A = X.mT @ X
+        B = b * A + c * (A @ A)
+        X = a * X + X @ B
+
+else: # Wide matrix (original math)
+    for a, b, c in polar_express_coeffs[:ns_steps]:
+        A = X @ X.mT
+        B = b * A + c * (A @ A)
+        X = a * X + B @ X
+
+return X
+```
+
+I had Claude write an XTX variant and verified that it was numerically equivalent.
+
+The speed improvement is likely too small to be measurable, but it's clear in the trace files.
+
+I came across it while exploring nanochat--here's the impact this fix has on a much larger (25-layer) model:
+
+## A100 Support
+
+I added a few things to make our code easier for anyone on a < H100 GPU.
+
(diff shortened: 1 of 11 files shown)


chrisjmccormick:
Leaving this as a draft for now because I'm not sure what will be needed exactly to make it a record, and also because it's time for sleep :)

ClassicLarry:
Maybe now that you have the shared optimizer you could check stepping the scalars/gates every step. Or maybe there is something around untieing the embeds for only the last X steps. A100 is nice!

chrisjmccormick:
I encountered something interesting while testing yesterday. Gathering statistics with 4 runs each, I tried removing 5 steps at a time from the `01------230` configuration. 

The results show:
1. Removing 5 steps had high variance and didn't hit the required `p`.
2. Removing 10 steps had low variance and hit it easily.
3. Removing 15 steps was slightly better than removing 10.
4. Removing an additional 10 (25 total) caused the average loss to jump by `+0.0024`

```
ve_12------012                  6 98.3442  0.0686    0.0000     NaN  3.2785  0.0013    0.0000 0.0184 - Baseline
ve_01------230                  0
ve_01------230  -5stps          4 98.7127  0.0728    0.3686  0.9999  3.2769  0.0021   -0.0016 0.0286 - Fail
ve_01------230 -10stps          4 98.3030  0.0388   -0.0412  0.1309  3.2770  0.0010   -0.0015 0.0052
ve_01------230 -15stps          4 98.0825  0.0896   -0.2617  0.0018  3.2769  0.0010   -0.0016 0.0040 - Best
ve_01------230 -25stps          4 97.4835  0.0551   -0.8607  0.0000  3.2793  0.0006    0.0008 0.0485 - Fail 
```

This fits well with frustrating experiences I've had recently, where a change seems to show strong improvement to loss over four runs, but then I try removing a modest number of steps and the loss jumps up. 

Here are a couple thoughts I had on what might be going on.

**Sensitivity to Batch Transition**

The model gets a big shock at the transition points, particularly at the 1/3s mark. I wonder if the contents of the training batches at and around this mark matter significantly to whether the model recovers well or not.

**Fine-Tuning Set**

At the end of training, during the end of cooldown and our extension window, I wonder if what we're doing there could be framed as fine-tuning on some training data in order to perform well on the validation set. 

Maybe the contents of that fine-tuning set are important for validation performance, and the wrong shift in the boundary could harm validation performance?

ClassicLarry:
I have also found there is an odd sensitivity going on. Your hypothesis sound plausible and could be tested by shuffling the documents. Another hypothesis is that there is something going on with the rounding on 1/3 and 2/3 of training, with respect to if it lands on a Muon only step vs a shared step, and how that lines up with when we are untying the embedding.

varunneal:
The underlying order of data matters definitely. I know this bc one time I had an error where I skipped the first shard and the losses were like 0.002-0.003 higher

chrisjmccormick:
Something kinda neat with the current baseline is that our 1/3 transition is at 512, 2/3s is at 1024, and total base steps is 1 shy of 1536. If those transition points seem to be good, I might try clamping them, and then adjusting the extension length, or the length of the final third if needed.

chrisjmccormick:
Here's an update on this. I ran some background experiments yesterday on a GH200 just to test loss.

Relative to the current record, simply tying the first and last VEs did increase loss (row 2 vs. row 1), so it's not an obvious better choice. 

Adding some extension steps improved the loss by -0.0028 (row 3 vs. row 2) so dialing the step count a bit might allow this configuration to come out ahead. I'll give it a try, but won't submit it unless it's somehow a decent marging.

Maybe a little more interesting is that tying the first and last table _and_ increasing the table count to 5 (like the current record) allows us to put that fifth table somewhere else. That seems like maybe a more meaningful direction to look at. Row 4 is the only arrangment I've tried so far.

```
                                            Runs   Time μ  Time σ  Time +/-  Time p  Loss μ  Loss σ  Loss +/-      p
ve_01------234 - 1575 stps  (baseline)         4 716.2268  1.5627    0.0000     NaN  3.2784  0.0015    0.0000 0.0632  
ve_01------230 - 1575 stps                     4 712.4075  0.6832   -3.8193  0.0052  3.2797  0.0011    0.0013 0.3062
ve_01------230 - 1584 stps                     4 718.7890  1.6491    2.5622  0.9675  3.2769  0.0009   -0.0015 0.0030
ve_0-1-2---340 - 1575 stps                     4 718.0358  0.9751    1.8090  0.9468  3.2778  0.0015   -0.0006 0.0312
```

On the added steps, what I did specifically was set it so the three training windows were 512 steps each (1536 total), and then added 8 extension steps (48 total), giving 1584 steps total. I'll try dropping it to +4 extension steps and running on an 8xH100 just to see how the timings come out.

photomz:
> The underlying order of data matters definitely. 

I've also noticed the same odd sensitivity, and it would worry me if our key findings are conditioned on data order. Maybe it would be prudent to instead take repeats with a varying dataloader order, s.t. when we test for significance, we know that it is a global inductive bias. Maybe I'll run some tests and see how severe this phenomenon is.

chrisjmccormick:
Closing this, though the polar express kernel did at least make it into my PR today. :)

PR #214, opened 2026-01-30
New Record: Mimetic V-O Initialization (-20 steps, -1.16s)

SulRash:
Changes

1. Mimetic V-O Initialization: initialize attention V and O projection weights with negative identity structure, idea taken from this paper by Trockman & Kolter (2023): https://arxiv.org/abs/2305.09828

```python
def mimetic_vo_init(W_V, W_O, num_heads, head_dim, beta=0.05, noise_std=0.01):
    """Initialize V and O matrices so W_V @ W_O.T ≈ -beta*I per head."""
    sqrt_beta = beta ** 0.5
    for h in range(num_heads):
        start, end = h * head_dim, (h + 1) * head_dim
        W_V.data[start:end, start:end] = sqrt_beta * torch.eye(head_dim)
        W_O.data[start:end, start:end] = -sqrt_beta * torch.eye(head_dim)
    # Noise for symmetry breaking
    W_V.data += noise_std * torch.randn_like(W_V)
    W_O.data += noise_std * torch.randn_like(W_O)
```

2. Reduced step count from 1575 to 1555

### Why it works

Standard transformer init zeros the output projections, making attention layers no-ops at init. Model wastes early steps learning to break this symmetry before attention contributes anything meaningful. Trockman & Kolter show that trained transformers naturally develop the `W_V @ W_O.T ≈ -βI` structure. Starting near this manifold skips wasted optimization steps.

The negative identity means attention outputs immediately subtract from the residual stream rather than doing nothing. This gives the model a head start on developing the typical "residual subtraction" pattern.

### What didn't work

- beta ≥ 0.10 was too aggressive, destabilizes training
- Mimetic initialization on Q-K showed no gain

### Timing and Validation

```python
import scipy.stats
import torch

# Record (1555 steps, mimetic init)
losses = [3.2775, 3.2791, 3.2805, 3.278, 3.2778, 3.2772, 3.2777]
times = [96.493, 96.586, 96.590, 96.535, 96.552, 96.556, 96.559]
# p=0.0036
# losses: (tensor(0.0011), tensor(3.2783))
# time: (tensor(0.0330), tensor(96.5530))

# Baseline (1575 steps, no mimetic init just retiming with my setup)
baseline_losses = [3.2759, 3.28, 3.2756, 3.2786, 3.2767, 3.2789, 3.2782]
baseline_times = [97.709, 97.617, 97.700, 97.726, 97.791, 97.705, 97.730]
print("baseline p=%.4f" % scipy.stats.ttest_1samp(baseline_losses, 3.28, alternative="less").pvalue)
# baseline p=0.0052
print("baseline losses:", torch.std_mean(torch.tensor(baseline_losses)))
# baseline losses: (tensor(0.0017), tensor(3.2777))
print("baseline time:", torch.std_mean(torch.tensor(baseline_times)))
# baseline time: (tensor(0.0520), tensor(97.7111))

# Baseline at 1555 steps (no mimetic init) - fails
baseline_reduced_losses = [3.2802, 3.2798, 3.2794, 3.2798, 3.2783, 3.2808, 3.2798, 3.2802, 3.2787, 3.28]
print("baseline reduced p=%.4f" % scipy.stats.ttest_1samp(baseline_reduced_losses, 3.28, alternative="less").pvalue)
# baseline reduced steps p=0.1149
```

Baseline at 1555 steps does not seem to reach significance, I think the init is doing real work.

Done with support from @bremen79

### References

Asher Trockman, & J. Zico Kolter. (2023). Mimetic Initialization of Self-Attention Layers. https://arxiv.org/abs/2305.09828

### DIFF
--- a/records/track_1_short/2026-01-30_MimeticValueOutput/runs/0f262f64-20c4-4268-9ae7-d7440c810abd.txt
+++ b/records/track_1_short/2026-01-30_MimeticValueOutput/runs/0f262f64-20c4-4268-9ae7-d7440c810abd.txt
@@ -1,0 +1,4131 @@
+import os
+import sys
+
+# Read the current file and the kernels file code ASAP, for logging
+with open(sys.argv[0], 'r') as f: 
+    code = f.read()
+with open(os.path.join(os.path.dirname(sys.argv[0]), 'triton_kernels.py'), 'r') as f:
+    code += f"\n\n{'-'*40}\n# triton_kernels.py\n{'-'*40}\n\n" 
+    code += f.read()
+
+import copy
+import glob
+import math
+import threading
+import time
+import uuid
+from dataclasses import dataclass
+from itertools import accumulate
+from pathlib import Path
+import gc
+
+os.environ["PYTORCH_ALLOC_CONF"] = "expandable_segments:True"
+import torch
+import triton
+
+torch.empty(
+    1, device=f"cuda:{os.environ['LOCAL_RANK']}", requires_grad=True
+).backward()  # prevents a bug on some systems
+import torch._dynamo as dynamo
+import torch.distributed as dist
+import torch.nn.functional as F
+
+# torch._inductor.config.coordinate_descent_tuning = True # we have banned this flag for new records because it causes compilation to take 30min
+from kernels import get_kernel
+from torch import Tensor, nn
+
+from triton_kernels import XXT, ba_plus_cAA, FusedLinearReLUSquareFunction, FusedSoftcappedCrossEntropy
+
+dynamo.config.recompile_limit = 64
+
+# -----------------------------------------------------------------------------
+# Custom operators: FP8 matmul by @YouJiacheng
+# Transposed layout by @ChrisJMcCormick allows for faster gradient accumulation.
+
+@torch.library.custom_op("nanogpt::mm_t", mutates_args=())
+def mm_t_op(x: Tensor, w: Tensor, x_s: float, w_s: float, grad_s: float) -> tuple[Tensor, Tensor, Tensor]:
+    """Computes y = x @ w with F8 weights stored as (in_features, out_features)."""
+    @torch.compile
+    def impl(x: Tensor, w: Tensor):
+        assert x.is_contiguous() and w.is_contiguous()
+        assert x.shape[1] == w.shape[0]  # x: (batch, in), w: (in, out)
+
+        x_f8 = x.div(x_s).to(torch.float8_e4m3fn)
+        w_f8 = w.div(w_s).to(torch.float8_e4m3fn)
+
+        # _scaled_mm requires column-major B. w_f8 is row-major (in, out).
+        # .T.contiguous().T creates a column-major view without changing logical shape.
+        w_f8_col_major = w_f8.T.contiguous().T
+
+        out = torch._scaled_mm(
+            x_f8,
+            w_f8_col_major,
+            out_dtype=torch.bfloat16,
+            scale_a=x.new_tensor(x_s, dtype=torch.float32),
+            scale_b=x.new_tensor(w_s, dtype=torch.float32),
+            use_fast_accum=True,
+        )
+        return out, x_f8, w_f8
+
+    return impl(x, w)
+
+@mm_t_op.register_fake
+def _(x: Tensor, w: Tensor, *_):
+    assert x.ndim == w.ndim == 2
+    assert x.shape[1] == w.shape[0]
+    assert x.device == w.device
+    assert x.is_contiguous() and w.is_contiguous()
+    return x @ w, x.to(torch.float8_e4m3fn), w.to(torch.float8_e4m3fn)
+
+@torch.library.custom_op("nanogpt::mm_t_backward", mutates_args=())
+def mm_t_backward_op(g: Tensor, x_f8: Tensor, w_f8: Tensor, x_s: float, w_s: float, grad_s: float) -> tuple[Tensor, Tensor]:
+    @torch.compile
+    def impl(grad: Tensor, x_f8: Tensor, w_f8: Tensor):
+        assert grad.is_contiguous()
+        
+        x_scale = grad.new_tensor(x_s, dtype=torch.float32)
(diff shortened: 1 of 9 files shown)


varunneal:
Nice saw a similar idea in this paper https://openreview.net/pdf?id=cKNOCYPo2W but it doesn't have symmetry breaking noise. You might want to look into semi orthogonal inits for q and k instead of random as proposed by this paper. 

Would it be possible to just bake in the initialization logic into the GPT class? might be able to use narrow + in place init, also don't need to duplicate the `with torch.no_grad():`

```
    for head_idx in range(self.num_heads):
        nn.init.uniform_(qkvo[0, head_idx])
        nn.init.uniform_(qkvo[1, head_idx])
        nn.init.eye_(qkvo[2, head_idx].narrow(-1, head_idx * self.head_dim, self.head_dim))  
        nn.init.eye_(qkvo[3, head_idx].narrow(-1, head_idx * self.head_dim, self.head_dim))
        # add noise at end
```

SulRash:
From a quick skim it looks like they benchmark against mimetic in that paper too, cool. I think with how saturated the record is initialization could be more valuable to explore right now. I tried semi-orthogonal on the MLP and it hurt performance when I combined it with mimetic V-O, I think it messes with Muon's orthogonalization? Not sure. Q/K might be different though, could be worth a shot later.

I don't think the indexing works here `narrow(-1, h*head_dim, head_dim)` grabs columns `(dim, head_dim)` but we need diagonal blocks `(head_dim, head_dim)` at `[start:end, start:end]`. I agree though it could be cleaner, maybe something like this:

```python
nn.init.uniform_(qkvo[0])
nn.init.uniform_(qkvo[1])
eye_block = torch.block_diag(*[torch.eye(self.head_dim) for _ in range(self.num_heads)])
qkvo[2].copy_(sqrt_beta * eye_block)
qkvo[3].copy_(-sqrt_beta * eye_block)
# add noise
```

ClassicLarry:
This looks interesting. Congrats on your execution- I had tried to implement this paper a while back and saw negative results. I think there may be a little bit more going on here on the 'why' side. The authors of that paper showed that the -beta*I relationship only occurred on vision transformers. On LLMs most layers were closer to the positive identity (which imo makes sense, we attend to tokens that we want to add to our representation, not subtract). 

Instead of looping over heads we can update the full diagonal at once
```
scaled_eye = 0.1**0.5 * torch.eye(model_dim, dtype=attn_bank.dtype, device=attn_bank.device)
attn_bank[:, 2*model_dim:3*model_dim] = scaled_eye
attn_bank[:, 3*model_dim:] = -scaled_eye
attn_bank[:, 2*model_dim:] += scale  * torch.randn_like(attn_bank[:, 2*model_dim:])
```

Can keep args on fineweb10B.

I am trying to think of why a negative identity might be good at initialization. One guess is that it functions as an 'activation dampener' early on in training. If say an MLP in layer 1 fires a massive activation for most positions into dimension 200. A positive identity pattern means that every attention layer will amplify this activation, so you get a sort of blowup down the layers anytime a dimension gets strongly activated. Negative identity has the opposite effect- the layer 2 attn would mute the massive activation by default. Though its not clear to me that muting the activation is actually preferred. Tracking W_V.T @ W_O.T over training might be informative. 

At initialization the 10 attention layers each function as a 10% dampening on the bos_token. I'm curious if this results in it serving as a much weaker attention sink early on in training, which sounds bad. But clearly something going on here is good.

SulRash:
Thanks! I ran the positive identity version and final val_loss was basically the same, the little difference could be attributed to noise. So the sign doesn't seem to matter here, just having non-zero structure in V-O at init. I also decided to track W_V.T @ W_O.T over training, that was a good suggestion:

| Init     | Diagonal sum (start to end) | Frob Norm (start to end) |
| -------- | --------------------------- | ------------------------ |
| Baseline | 0.00 to -7.38               | 0.00 to 353.0            |
| Negative | -38.45 to -50.10            | 3.52 to 347.0            |
| Positive | +38.30 to +163.7            | 3.51 to 355.4            |

In the diagonal sum positive grows, negative stays put, and baseline drifts slightly negative but total magnitude ends up the same, all three reach similar final frob norms and the model develops V-O structure naturally during training independent of init. Sign is preserved (negative stays negative, positive grows more positive), but doesn't seem to affect final performance.

So I don't think it's the activation dampening mechanism since baseline develops similar magnitude V-O structure without any identity init. More likely it's just that giving attention non-zero structure at init lets it contribute meaningfully from step 1 instead of being effectively disabled until the weights grow? bos_token sink point is interesting too. The 10% dampening per layer would compound to like 35% of the original by the end. Not sure if weaker sink is good or bad. Might actually help if it forces the model to use real tokens for attention earlier in training instead of relying on the sink.

_EDIT: The code you suggested is cleaner and works well as a replacement_

ClassicLarry:
Cool that experiment is informative. Continuing that direction, looks like we can recover the full gain with a simple:
```
# improved init scale by @YouJiacheng and @SulRash
std = 0.5 * model_dim ** -0.5
bound = (3 ** 0.5) * std
with torch.no_grad():
    self.attn_bank.uniform_(-bound, bound)
    self.mlp_bank[:, 0, :, :].uniform_(-bound, bound)  # c_fc
    self.mlp_bank[:, 1, :, :].zero_()  # c_proj - zero init suggested by @Grad62304977
```
losses =[3.2773, 3.2772, 3.2767, 3.2786]
p=0.0041
This roughly replicates the amount of noise you have added.

The way the identity is being applied makes it so head1 only applies the identity for residual dims [0:128], then head 2 handles the identity for dims [128:256], etc. This means that each head isn't really doing a full identity op of the token it attends to, just a subset of dims. As you mention, I think this is kind of a wash in the noise.

This is typically how I find things go where a initial hypothesis ends up being useful not because of its direct theory, but because it motivates experiments that uncover other things.

I will merge this as-is so you are added as a contributor, and then do a follow up merge to simplify the init and correct the data path.

ClassicLarry:
Timed at -1.2 so 99.0-1.2 = 97.8s = 1.63 minutes.

PR #215, opened 2026-01-30
[New Record] Single weight for value embeddings + refactors (-1s)

varunneal:
This record combines the value embeddings into a single weight:

```
-        ve = [value_embed(input_seq) for value_embed in self.value_embeds]
+        ve = self.value_embeds.view(5, self.vocab_size, -1)[:, input_seq]
```

and fixes the step 5 blocking behavior from the async data loader.

Additionally, there are quite a few refactors and that don't change the timing but decrease LOC and hopefully improve quality of life:

* Froze/hardcoded the best Triton configs for the symmetric matmul kernels (see [discussion](https://github.com/KellerJordan/modded-nanogpt/discussions/23?sort=new#discussioncomment-15532866)). Warmup should be faster now.
* Merged the Yarn and YarnPairedHead classes, and the CausalSelfAttention and PairedHeadCausalSelfAttention classes, mostly for readability and LOC reasons
* The Training Schedule has been refactored to use a `TrainingSchedule` class. The separate batch size, window size, MTP, and learning rate schedules have been unified. Each `TrainingStage` corresponds to a torch compile graph.
* Added a `grad_scale`. The major idea here is that the gradient magnitude should match between 1 and 8 GPUs for training consistency (matters for FP8 scales especially, but also BF16 a little bit). I noticed a `grad_scale = 2` is slightly better than `grad_scale = 1` (default), though not enough to warrant lowering steps.
* Moved the distributed setup to the top of the file, and also put Hyperparameters earlier in the file.
* Cleaned up Hyperparameters.

Some of the refactors are quite subjective, so happy to edit to incorporate others' feedback.

## Timing and Validation

This PR improves the step time and does not impact (or very slightly decreases) loss.

```
import scipy.stats
import torch

losses = [3.2767, 3.2777, 3.2795, 3.2774, 3.2799, 3.2779, 3.2777]
times = [97.825, 97.777, 97.818, 97.777, 97.745, 97.715, 97.728]

# p=0.0025

# losses: (std=0.0012, mean=3.2781)

# time: (std=0.0425, mean=97.7693)
```

```
import scipy.stats
import torch

losses = [3.2776, 3.2803, 3.2773, 3.2794, 3.2784, 3.2794]
times = [98.908, 98.878, 98.796, 98.667, 98.800, 98.776]

# losses: (std=0.0012, mean=3.2787)

# time: (std=0.0848, mean=98.8042)
```

These timings show $\approx 1$ seconds of improvement.

## Discussion

I mostly want to push out this PR because there is a lot of ongoing effort being put into adding more value embeddings. It turns out that the "fused" value embedding saves a considerable amount of time —— something like 0.8s throughout training. This will probably allow us to justify more value embeds. Note that the current approach is basically what was done earlier in the speedrun, but it was replaced due to speedups in the optimizer step when separating each value embed (record 16). Due to the new optimizer architecture, this previous consideration is no longer there, and the speedups in the lookup matter quite a bit. By my count, we are saving ~10 reads of the input sequence between the forward and backward step.

I created a new class `Shard` to replace the previous two-class BOSFinder and AsyncDataLoader. It is my understanding that fixing the step 5 blocking behavior saves ~100ms.

I believe there is around 0.3s of potential improvement in the current data loader by overlapping the CPU work of generating the input tensors (`cat`s and indexing) and the H2D CPU -> GPU transfer. I was able to get something like this working by adding a hook in the optimizer step, but my approach was a bit messy. Going to put off working on this for now.

### DIFF
--- a/README.md
+++ b/README.md
@@ -48,7 +48,7 @@ Contributors list (growing with each new record): [@bozavlado](https://x.com/boz
 [@akash5474](https://github.com/akash5474), [@snimu](https://x.com/omouamoua), [@roeeshenberg](https://x.com/roeeshenberg),
 [@ChrisJMcCormick](https://x.com/ChrisJMcCormick), [@dominikkallusky](https://github.com/dominikkallusky), [@acutkosky](https://github.com/acutkosky), 
 [@manikbhandari](https://github.com/manikbhandari), [@andrewbriand](https://github.com/andrewbriand), [@jrauvola](https://github.com/jrauvola),
-[@soren_dunn_](https://x.com/soren_dunn_), [@photon_mz](https://x.com/photon_mz)
+[@soren_dunn_](https://x.com/soren_dunn_), [@photon_mz](https://x.com/photon_mz), [@srashedll](https://x.com/srashedll)
 
 
 ---
@@ -165,6 +165,8 @@ Note: The 3.28 target was selected to match [Andrej Karpathy's GPT-2 (small) rep
 61 | 1.748 minutes | [Unified Optimizers and Transposed LM Head](https://x.com/classiclarryd/status/2013399457841160702) | 01/18/26 | [log](records/track_1_short/2026-01-18_UnifiedOptimizers/unified-optimizer/2fc79469-a527-4bde-8540-8426ed3352d1.txt),[PR](https://github.com/KellerJordan/modded-nanogpt/pull/200) | @ChrisJMcCormick
 62 | 1.655 minutes | [Bigram Hash Embedding](https://x.com/classiclarryd/status/2013520088297558274) | 01/19/26 | [log](records/track_1_short/2026-01-19_BigramHashEmbedding/40ec7bb6-14b3-46f8-90b7-bb5ed188faba.txt),[PR](https://github.com/KellerJordan/modded-nanogpt/pull/201) | @classiclarryd
 63 | 1.650 minutes | [Untie Value Embeds](https://x.com/classiclarryd/status/2016968386476200301) | 01/26/26 | [log](records/track_1_short/2026-01-26-UntieValueEmbeddings/43955d93-6914-40cb-bdf8-786ace93784f.txt),[PR](https://github.com/KellerJordan/modded-nanogpt/pull/209) | @photo_mz
+64 | 1.630 minutes | Tuned nonzero Attn V and O init | 01/30/26 | [log](records/track_1_short/2026-01-30_MimeticValueOutput/runs/0f262f64-20c4-4268-9ae7-d7440c810abd.txt),[PR]([https://github.com/KellerJordan/modded-nanogpt/pull/209](https://github.com/KellerJordan/modded-nanogpt/pull/214)) | @srashedll
+
 ## Rules
 
 New records must:
--- a/records/track_1_short/2026-01-30_MimeticValueOutput/runs/0f262f64-20c4-4268-9ae7-d7440c810abd.txt
+++ b/records/track_1_short/2026-01-30_MimeticValueOutput/runs/0f262f64-20c4-4268-9ae7-d7440c810abd.txt
@@ -1,0 +1,4131 @@
+import os
+import sys
+
+# Read the current file and the kernels file code ASAP, for logging
+with open(sys.argv[0], 'r') as f: 
+    code = f.read()
+with open(os.path.join(os.path.dirname(sys.argv[0]), 'triton_kernels.py'), 'r') as f:
+    code += f"\n\n{'-'*40}\n# triton_kernels.py\n{'-'*40}\n\n" 
+    code += f.read()
+
+import copy
+import glob
+import math
+import threading
+import time
+import uuid
+from dataclasses import dataclass
+from itertools import accumulate
+from pathlib import Path
+import gc
+
+os.environ["PYTORCH_ALLOC_CONF"] = "expandable_segments:True"
+import torch
+import triton
+
+torch.empty(
+    1, device=f"cuda:{os.environ['LOCAL_RANK']}", requires_grad=True
+).backward()  # prevents a bug on some systems
+import torch._dynamo as dynamo
+import torch.distributed as dist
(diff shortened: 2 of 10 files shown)


ClassicLarry:
Merged, will update readme at -1s later. I am uncertain if the 2x grad is actually better now that we are down 20 steps with diff attn init, but still seemed below loss cutoff.

ClassicLarry:
I ran some tests to isolate the speedup. Your method of querying the value embeds is 25% faster on H100. Interestingly, its slower on A100.
Summary:
```
vocab_size = 50304
model_dim = 768

# 0.003s per fwd/bwk pass
self.value_embeds = nn.ModuleList([nn.Embedding(vocab_size, model_dim, dtype=torch.bfloat16) for _ in range(5)])
ve = [value_embed(input_seq) for value_embed in self.value_embeds]

# 0.0023s per fwd/bwk pass
self.value_embeds = nn.Parameter(torch.rand(vocab_size*5, model_dim, dtype=torch.bfloat16))
ve = self.value_embeds.view(5, 50304, -1)[:, input_seq]

# (0.003-0.0023)*1500steps = ~1s improvement
```


```
x = 1
import torch
import torch.nn as nn


class Model1(nn.Module):
  def __init__(self, vocab_size=50304, model_dim=768):
    super().__init__()
    self.embed = nn.Embedding(vocab_size, model_dim, dtype=torch.bfloat16)
    self.value_embeds = nn.Embedding(vocab_size, model_dim*5, dtype=torch.bfloat16)
    self.lambdas = nn.Parameter(torch.ones(5, dtype=torch.bfloat16))
  def forward(self, input_seq):
    ve = self.value_embeds(input_seq).view(5, -1)[:, ]
    ve = [ve[0], ve[1], ve[2], ve[3], ve[4]]
    x = self.embed(input_seq)
    for i in range(5):
      x = x + self.lambdas[i]*ve[i].view_as(x)
    loss = x.sum()
    return loss

class Model2(nn.Module):
  def __init__(self, vocab_size=50304, model_dim=768):
    super().__init__()
    self.embed = nn.Embedding(vocab_size, model_dim, dtype=torch.bfloat16)
    self.value_embeds = nn.ModuleList([nn.Embedding(vocab_size, model_dim, dtype=torch.bfloat16) for _ in range(5)])
    self.lambdas = nn.Parameter(torch.ones(5, dtype=torch.bfloat16))
  def forward(self, input_seq):
    ve = [value_embed(input_seq) for value_embed in self.value_embeds]
    ve = [ve[0], ve[1], ve[2], ve[3], ve[4]]
    x = self.embed(input_seq)
    for i in range(5):
      x = x + self.lambdas[i]*ve[i].view_as(x)
    loss = x.sum()
    return loss

class Model3(nn.Module):
  def __init__(self, vocab_size=50304, model_dim=768):
    super().__init__()
    self.embed = nn.Embedding(vocab_size, model_dim, dtype=torch.bfloat16)
    self.value_embeds = nn.Parameter(torch.rand(vocab_size*5, model_dim, dtype=torch.bfloat16))
    self.lambdas = nn.Parameter(torch.ones(5, dtype=torch.bfloat16))
  def forward(self, input_seq):
    ve = self.value_embeds.view(5, 50304, -1)[:, input_seq]
    ve = [ve[0], ve[1], ve[2], ve[3], ve[4]]
    x = self.embed(input_seq)
    for i in range(5):
      x = x + self.lambdas[i]*ve[i].view_as(x)
    loss = x.sum()
    return loss

class Model4(nn.Module):
  def __init__(self, vocab_size=50304, model_dim=768):
    super().__init__()
    self.embed = nn.Parameter(torch.rand(vocab_size, model_dim, dtype=torch.bfloat16))
    self.value_embeds = nn.Parameter(torch.rand(vocab_size*5, model_dim, dtype=torch.bfloat16))
    self.lambdas = nn.Parameter(torch.ones(5, dtype=torch.bfloat16))
  def forward(self, input_seq):
    ve = self.value_embeds.view(5, 50304, -1)[:, input_seq]
    ve = [ve[0], ve[1], ve[2], ve[3], ve[4]]
    x = self.embed[input_seq]
    
    for i in range(5):
      x = x + self.lambdas[i]*ve[i].view_as(x)
    loss = x.sum()
    return loss

data = torch.randint(50304, (1, 2048*16), device='cuda')
model1 = Model1().cuda()
model1 = torch.compile(model1, dynamic=False, fullgraph=True)

model2 = Model2().cuda()
model2 = torch.compile(model2, dynamic=False, fullgraph=True)

model3 = Model3().cuda()
model3 = torch.compile(model3, dynamic=False, fullgraph=True)

model4 = Model4().cuda()
model4 = torch.compile(model4, dynamic=False, fullgraph=True)

import time

def test(model):
  torch.cuda.synchronize()
  t0 = time.perf_counter()
  for _ in range(1000):
    model(data).backward()
    model.zero_grad(set_to_none=True)
  torch.cuda.synchronize()
  run_time = time.perf_counter() - t0
  print(run_time)

print('try')
test(model1)
test(model2)
test(model3)
test(model4)

print('try')
test(model1)
test(model2)
test(model3)
test(model4)

print('try')
test(model1)
test(model2)
test(model3)
test(model4)

print('try')
test(model1)
test(model2)
test(model3)
test(model4)

print('try')
test(model1)
test(model2)
test(model3)
test(model4)
```

karpathy:
The ve fusion was easy to try in nanochat (it's a great idea and it simplifies a bunch of the surrounding code too) but it turned out *slower* by about 1% (in tok/s) for a reason I don't understand, weird.

varunneal:
@karpathy It may have to do with how your optimizer is set up (e.g. waiting on a big value embed weight instead of overlapping communicating and processing each weight). If it's possible to isolate, maybe try timing forward/backward without an optimizer step.

karpathy:
Yeah I was looking at optim as a culprit too. The AdamW compute is basically instantaneous though, so it doesn't seem like there'd be much to overlap. That said, the drop is only 1% too. I'll probably come around to profile it a bit more.

PR #219, opened 2026-02-03
Sparse comms for bigram-embeds

shenberg:
Finally got a non-trivial improvement by using sparse communication (roughly 1 second, but I'm based on a slightly old branch, will update to newer, which will probably reduce the improvement by a bit). There is still some room for improvement, however. The code is pretty raw (hard-coded some extra bookkeeping and checks for world_size=8) and I'm not quite sure what's the correct place for all of the changes. Some is pretty self-explanatory (adding a `sharded_sparse` mode to comms), but in order to achieve maximum overlap, the training loop was intrusively modified. Some advice would be more than welcome.

The outline of the strategy:
Each rank is owns (=responsible for updating) 1/8th of the bigram embedding table, and needs to send to each rank the local gradient rows that have been populated by bigrams it saw, and likewise needs to receive rows it owns from local gradients of other ranks. 
Example: 
We have two ranks, 0, 1 and our bigram list is of size 4: [0,1,2,3]
So rank 0 owns bigrams 0,1 and rank 1 owns bigrams 2,3
Rank 0 batch: [1,2]
Rank 1 batch: [1,3]
So rank 0 needs to send to rank 1 its local gradient row 2, and receive row 1.
Rank 1 needs to send gradient row 1 to rank 0, and receive gradient row 2.

The strategy we use is:
0. Since this is only on Adam steps, we cache the previous muon-only step's bigrams on the CPU
1. Before the forward pass, each rank shares with the other how many gradient rows it's going to send them. This is done async on the CPU and overlaps with the forward pass
2. After forward, before backwards, we share the actual indexes gradient indexes we're going to send between ranks (GPU, async, overlaps with start of backward)
3. in the scatter phase, we send between ranks the updated rows only (all_to_all_single), and then on each rank we sum those rows into our gradient
4. gather is still dense since our optimizer is dense

Extra notes:
I did some work around pinning in the loader since this was slowing down the main thread with the extra CPU work added. Maybe just this can be back-ported to master to get most of the improvement. Not sure.
In addition, after staring at some profiles, I did a slight re-order of scatter and gather orders (`param_table` and `work_order` respectively) which seemed to improve overlap. Again, possibly part of the gain comes from it irrespective of any changes related to sparse communication.

Most of the hard work is done before forward(): we get a sorted unique list of bigram indexes since the last optimizer step, then find the boundaries between ranks in this list in order to get the list of gradient rows we need to send each rank. It's optimized enough at the moment to overlap with the previous step on the GPU.

Abandoned approaches:
1. Do all the work on the GPU: unfortunately `all_to_all_single()` only supports split sizes on the CPU, so every method without padding required a cpu<->gpu sync. Device-initiated comms are alpha (symmetric memory) but should theoretically enable this.
2. Work in a background thread or even just a different cuda stream: I moved the cpu<->gpu sync to a background thread, however it's very problematic to use nccl from two threads at the same time. Turns out that the various ranks are very much not in sync so I would get different orders of entering into collectives. Even with two streams on the same thread execution order isn't guaranteed across ranks.

### DIFF
--- a/train_gpt.py
+++ b/train_gpt.py
@@ -218,11 +218,11 @@ def sparse_comms_start(idxes_np, N, rank, world):
     )
     send_counts = torch.from_numpy(insertion_points[1:] - insertion_points[:-1])
     # zero-out own send-count - we won't send our own gradient rows to ourselves as it's a waste:
-    # in the optimizer, we'll use the slice of the gradient that already includes them as the basis
+    # in the optimizer, we'll use the slice of the gradient that already includes them as the base tensor
     send_counts[rank] = 0
+
     # remove our own indexes from the send list
     send_idxes = torch.cat([send_idxes[: insertion_points[rank]], send_idxes[insertion_points[rank + 1] :]])
-    # print(f"{rank = }, {send_counts = }, {send_counts.sum() = }, {send_idxes.shape = }", flush=True)
 
     # share the send counts so that each rank will know how many rows
     # to expect from every other rank
@@ -281,8 +281,6 @@ def sparse_comms_merge_gradients(grad, recv_idx, recv_vals, rank, world):
     d = grad.shape[1]
     rows_per_rank = grad.shape[0] // world
 
-    # grad_slice = grad[rows_per_rank * rank : rows_per_rank * (rank + 1)] * (1 / world)
-    # grad_slice.index_add_(0, recv_idx.remainder_(), recv_vals.view(-1, d), alpha = (1 / world))
     grad.index_add_(0, recv_idx, recv_vals.view(-1, d))
 
     return grad[rows_per_rank * rank : rows_per_rank * (rank + 1)].mul_((1 / world))
@@ -1956,7 +1954,7 @@ for step in warmup_steps:
         send_args = training_manager.train_loader_send_args
         inputs, targets, cum_seqlens, bigram_inputs, bigram_cpu = train_loader.send(send_args)
 
-        # TODO: hack
+        # TODO: cleaner integration of sparse comms
         if world_size == 8:
             if not training_manager._is_adam_step(step):
                 bigrams_old = bigram_cpu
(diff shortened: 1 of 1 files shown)


photomz:
Amazing work! I was testing a version of this but couldn't get the implementation to be as efficient.
1. How much buffer is there between the CPU finishing computation and the start of bwd? Are there arch changes we should be mindful of to not create race contentions / break the latency overlap.
2. Can this work for lower sparsity ratios, ie more params activated per device bz, such as in the VE? I recall VE activation was ~70% at max final bz (lower than a uniform pb calculation due to the inefficient tokenizer).
3. What would it take to make gather sparse, perhaps we can take low-rank appx of Adam like in Adafactor, then we'd only need to gather the params with nonzero grad?

shenberg:
1. It's a bit hard for me to tell: When I profile this, the profiler overhead makes this behave strangely - over 10 iterations the gloo collective increases from taking 4ms to taking ~50ms, I believe because only rank0 is paying the profiling price. I don't know what's happening without the profiler. 
eg. early-profile step (biggest batch-size):
image

Late-profile step:
image


2. I started thinking of this in the context of VE, I think it's a no, but not necessarily for the comms price, rather because the kernel to reconstitute the dense gradient is pretty expensive, and it's hidden inside the comms when you use reduce_scatter. Right now I overlap it with the reduce-scatter of everything else, if we did all of them this way we'd run out of comms to overlap with computation and it'd probably take longer overall: the price of the kernel to merge the gradients should roughly scale with the amount of rows we need to merge, so it probably wouldn't be much faster for a VE than for the bigram embedding, while the comms savings would be much smaller.

Here's a somewhat-annotated profile of the timings:
image

3. PyTorch actually already has an implementation called "SparseAdam" that basically only updates the momentum and variance on rows that have seen use this iteration, but this is intended for pytorch-style sparse gradients (what you get if you do `nn.Embedding(sparse=True)`), which are unsupported in collectives at the moment (and the solution I found on the interwebs requires two device->host syncs).

Because I thought it's "wrong" to update optimizer state for vectors we didn't see in the batch at all, I tried in the past to implement a sparse update for embeddings and VE and it worked terribly. Theory 1 that I had was that I just didn't tune my implementation enough (maybe!). My favorite theory, though, is that what's happening is that rare tokens get their learning rate increased by rarity: if most of the time the gradient is 0, the variance shrinks. Since momentum (beta1) is relatively small, when we do get a non-zero gradient, the update will be big due to the divide by variance. Maybe this is a good thing.

Re: Adafactor, from what I understand, it'd save space on variance storage, but parameter updates would still be dense. The paper however does imply that what I suggested as good (too-low variance estimate causing large updates) is in fact harmful.

shenberg:
Cleaned up version based on newer commit at #221

PR #216, opened 2026-01-31
[New Record] Removing H2D Movement in Bigram Hash Embedding (-0.25s)

dhruvji:
This simple optimization moves the bigram hash computation to gpu. It effectively eliminates one H2D transfer per batch.

As such, it also simplifies data loader and makes it so there is one fewer tensor to manage in the async pipeline.

## Timing and Validation
This PR improves the time and does not impact loss on my hardware, tested in Docker.
```python
import scipy.stats
import torch

# First set (New Record group)
losses1 = [3.2771, 3.2781, 3.2787, 3.2786, 3.2761, 3.2777, 3.2783]
times1 = [97887, 97942, 98031, 97906, 97840, 97901, 97843] # in ms

# Second set (Prev Record group)
losses2 = [3.2778, 3.2780, 3.2777, 3.2818, 3.2769, 3.2780]
times2 = [98036, 98178, 98158, 98135, 98137, 98146] # in ms

print("=== New Record Set ===")
print("p=%.4f" % scipy.stats.ttest_1samp(losses1, 3.28, alternative="less").pvalue)
# p=0.0004
print("losses:", torch.std_mean(torch.tensor(losses1)))
# losses: (tensor(0.0009), tensor(3.2778))
print("time:", torch.std_mean(torch.tensor(times1, dtype=torch.float32)))
# time: (tensor(65.3489), tensor(97907.1406))

print("\n=== Prev Record Set ===")
print("losses:", torch.std_mean(torch.tensor(losses2)))
# losses: (tensor(0.0017), tensor(3.2784))
print("time:", torch.std_mean(torch.tensor(times2, dtype=torch.float32)))
# time: (tensor(49.4719), tensor(98131.6641))

``` 
## Discussion
I did want to clarify whether my `torch.compile` usage is completely okay (the flags in the decorator are in other decorators). This tiny patch was largely inspired during my first proper pass on the nanogpt repo, which I found through the Bigram Hash Embedding tweet. Excited to contribute going forward!

PR #221, opened 2026-02-06
New WR: sparse bigram gradient comms (-0.6 seconds)

shenberg:
Title timing is a mistake, its closer to -0.75 seconds. 

This is an update & cleanup of #219. No ML changes.

This PR contains three changes:
1. Sparse comms for bigram embeddings
2. Moved the bigram index calculation to write directly into a pinned tensor
3. Changed scatter/gather order a bit to improve overlap

The sparse comms implementation saves much bandwidth (even at largest batch size, the amount of communication is about the same as the embedding and lm_head layers (for the scatter)) at the cost of extra compute to reconstitute the rank-local gradient, which is hidden by overlap with other communication.

Moving the bigram index calculation directly into a pinned tensor saved a lot of time in the forward pass as `.to(device, non_blocking=True)` was very slow. Since I need the index on the CPU as part of the sparse communication scheme, this is mutually exclusive with #216 unfortunately, though I think much of the gain in #216 is already folded into moving to a pinned tensor.

Ablation: a build with only changes 2 & 3 was about 500ms faster than baseline (hard-coded `_sparse_comms_active()` to `return False`).

The scatter-gather order is a bit under-explored: using a profiler, it's easy to see that overlap is not perfect despite there being enough transfers queued. I found a configuration that improved timing by another 100ms (there should be about an extra 200ms on top of it) but for some reason loss was increased.

Will upload the logs in a bit. See #219 for more details of the sparse communication algorithm until I move them here to the readme

### DIFF
--- a/README.md
+++ b/README.md
@@ -47,8 +47,8 @@ Contributors list (growing with each new record): [@bozavlado](https://x.com/boz
 [@bernard24](https://github.com/bernard24)/https://www.hiverge.ai/, [@Gusarich](https://x.com/Gusarich), [@li_zichong](https://x.com/li_zichong),
 [@akash5474](https://github.com/akash5474), [@snimu](https://x.com/omouamoua), [@roeeshenberg](https://x.com/roeeshenberg),
 [@ChrisJMcCormick](https://x.com/ChrisJMcCormick), [@dominikkallusky](https://github.com/dominikkallusky), [@acutkosky](https://github.com/acutkosky), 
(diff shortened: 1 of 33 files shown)


chrisjmccormick:
FYI, in case it's helpful for catching things up--I integrated the PRs prior to yours into your code, and re-ran it for four runs.
The updated code and logs are in a subfolder of my PR #233.

ClassicLarry:
Merging at 91.0 (-0.3s) based on re-timing. (This one undoes the 0.3s gain from moving the bigram to GPU)

I will do a merge afterwards to clean up the merge conflicts.

PR #224, opened 2026-02-10
New Record: Method (LR Floor ↑, Short-Sequence Curriculum, Short Window) and Paper Generated by AI Agents (–0.8s)

dualverse-ai:
We present a straightforward improvement over current SOTA [#207](https://github.com/KellerJordan/modded-nanogpt/pull/207) discovered by our agent system, **The Station** (https://github.com/dualverse-ai/station). The key modifications are summarized below:

- _`L1500-L1502` of `train_gpt.py`_: LR floor increase (0.1 → 0.15) + denominator guard  
    - Contribution: higher effective lr during cooldown, more stable late-stage convergence, avoids “stalling”.
- _`L1550, L1632-L1633` of `train_gpt.py`_: Stage3 forced short-window (ws_short=2)
    - Contribution: reduces attention compute; directly updates TrainingManager.ws_short.
- _`L1547-L1548, L1626-L1629` of `train_gpt.py`_: Stage1 short-sequence curriculum (2048 → 896 → 2048) and _`L1645-L1648`_: Forced DataLoader refresh (step=1)
    - Contribution: significantly lower compute early on, shorter overall training time, while restoring full length later to preserve quality. And Forced DataLoader refresh ensures the seq_len change is applied immediately at step 1, avoiding schedule desync.

**Performance statistics** were collected on 3 machines, with the baseline SOTA [#207](https://github.com/KellerJordan/modded-nanogpt/pull/207)  shown on the left, and our implementation of three modifications (denoted as Eval_248 in our Station) shown on the right. Each machine reports 5 runs, for a total of 15 runs:

e66f895a8149bb66c58dbd21a71051f9

```
>>> from scipy import stats
>>> losses = [3.2779,3.2788,3.2791,3.2752,3.2752,3.2784,3.2779,3.2784,3.2777,3.278,3.2804,3.2806,3.2786,3.2777,3.2776]
>>> print("p=%.4f" % stats.ttest_1samp(losses, 3.28, alternative="less").pvalue)
p=0.0001
```

**Ablation on three modifications**:

We used a single node to run all ablations, along with the baseline SOTA [#207](https://github.com/KellerJordan/modded-nanogpt/pull/207) and full method (Eval_248). `Ablation w/ LR↑` means that this ablation includes only the LR↑ modification. 

*Note LR↑ helps stabilize convergence, while the other two modifications (short-sequence and short-window) should help accelerate training.*

5be39f0b069fe0f2f74615a0be3ccb7f

b53a165c6825b158e23e15da1eae254a

Many thanks for providing and maintaining such a great platform!

**Note1:** Because of cluster limts, our machines are unable to run Docker. Therefore, we replicated the environment and tested the method on several machines to assess statistical significance. While all algorithms run about 1–2 seconds slower on our hardware, the relative performance differences are still clearly noticeable.

**Note2:** This improvement (Eval 248) builds on the accumulated experience from the entire [Station](https://github.com/dualverse-ai/station) session. As agents are encouraged to write papers to document their progress, below is a short paper produced by the same agent who implemented Eval 248:

(*Note that the speed reported in the paper is for reference only, as it is calibrated against the leaderboard score to ensure consistency across different nodes.*)

## Abstract: 
We show that a common “stage3 window shrink” speedup for `baseline_ImprovedLMHead_93` can silently become a no-op if implemented by only editing `args.window_size_*`, because the baseline
  attention forward reads `TrainingManager.ws_short/ws_long` (runtime state) instead. Applying the shrink
  correctly at boundary-aligned stage transitions yields a best observed 91.92s mean-of-3 (Eval 248,
  final mean loss 3.2783) and a more robust 92.38s mean-of-3 via a global LR floor clamp (Eval 262, final
  mean loss 3.2788). We provide ablations that isolate the ws_short effect (~0.25s) and characterize variance
  cliffs across nodes.

## Introduction

Recent Station progress on Task 1 relies on *boundary-aligned* compute reductions (sequence-length warmup, attention window shrinking, LR-floor shaping) layered on top of the system `baseline_ImprovedLMHead_93`. A recurring practical failure mode is that wrapper-based experiments can apply an intended intervention at the wrong *control point* (e.g., editing an argument that is not actually consulted by the compiled forward path). This wastes evaluations and makes it hard to interpret ablations.

This paper identifies such a mismatch for “stage3 attention window shrink”, fixes it in a policy-safe way (no loss/logit manipulation), and quantifies the speed/margin trade-offs.

### Baseline

All experiments are wrappers around `storage/system/baseline_ImprovedLMHead_93.py` (NorMuon+Adam schedule, fused FP8 head + fused CE, multi-stage batch/window schedules).

**Key mechanistic detail (root cause):**
the baseline attention implementation reads *runtime* fields `TrainingManager.ws_short` and `TrainingManager.ws_long` inside the forward path, rather than reading `args.window_size_*`. Therefore:

- modifying only `args.window_size_*` may have **zero effect** (silent no-op),
- the correct control point for window shrinking is mutating `TrainingManager.ws_short/ws_long` at the relevant steps.

### Wrapper Hygiene (to avoid non-scientific failures)

When loading the system baseline via `importlib`:

- register the module in `sys.modules` before `exec_module` (required for decorators/dataclasses),
- propagate `world_size` into the loaded module (`setattr(mod, "world_size", world_size)`) because the compiled forward reads module-global `world_size` on some paths,
- define literal `def set_context` / `class Hyperparameters` / `class TrainingManager` in `submission.py` (static validator can reject “assignment-only” wrappers).

These are implementation constraints, not scientific contributions, but are necessary to reproduce the experiments.

### Boundary-Aligned Recipe Components

All successful lines here use the same stage boundaries as the baseline schedule:

- Stage1: steps `1..504`
- Stage2: steps `505..1009`
- Stage3: steps `1010..1514`
- Step `1515` is the next boundary (stage4/extension logic varies by baseline)

**Stage1 short-sequence warmup (Seq896):**
set `args.train_max_seq_len = 896` for steps `1..504` and revert to `2048` at step `505`. Importantly, set the value **before** calling the baseline `advance_schedule(step)` so any dataloader-send inside the baseline uses the intended seq-len.

**Stage3 short-window shrink (WSshort2):**
apply `TrainingManager.ws_short = 2` only for `1010 <= step < 1515`. We intentionally leave `ws_long` baseline-controlled unless explicitly stated.

**LR floor shaping:**
two schedules are studied:

- **Cooldown interpolation**: linearly interpolate towards `lr_floor=0.15` only during cooldown (used in Eval 248; fastest observed but cliffy).
- **Global clamp**: `lr(step) = max(lr_baseline(step), 0.15)` for *all* steps (used in Eval 262; more robust cross-node).

## Results

Station metric is mean time-to-target across 3 runs; runs that do not reach the target receive a 999999s penalty. We compute the reported `±` as the standard deviation of the three per-run `train_time` values shown in the logs.

| Eval | Key Diffs vs Baseline | Time-to-Target (s) | Final Mean Val Loss | Status |
|---:|---|---:|---:|---|
| Eval 18 | +5 extension steps (minimal pass baseline early on) | 93.761 ± 0.232 | 3.277933 | PASS |
| Eval 248 | **Fix control point**: Stage3 `ws_short=2` via `TrainingManager.ws_short`; Seq896; cooldown-interp to floor=0.15; ext=40 | **91.922 ± 0.313** | 3.278333 | PASS |
| Eval 262 | Same as 248 but **global LR clamp** floor=0.15 (no cooldown interpolation) | 92.380 ± 0.484 | 3.278767 | PASS |
| Eval 277 | **Ablation**: remove stage3 ws_short shrink; keep global clamp + Seq896 + ext=40 | 92.631 ± 0.038 | 3.275767 | PASS |
| Eval 276 | Add mild stage2 seq cut (1920 for steps 505..1009) on top of global clamp + Seq896 + WSshort2 | 92.133 ± 0.297 | 3.278933 | PASS |
| Eval 267 | Stage2 seq=1536 under global clamp family | 999999 (target not reached) | best 3.279400 | FAIL |
| Eval 272 | Stage2 seq=1664 under global clamp family | 999999 (target not reached) | best 3.279300 | FAIL |

**Isolating ws_short effect (single-diff comparison):**
comparing Eval 262 vs Eval 277 shows that setting `ws_short=2` in stage3 saves ~0.25s on this node-calibrated metric, but reduces margin (final loss moves closer to 3.279).

## Discussion

**Control-point alignment matters:** the largest *qualitative* contribution is preventing a silent no-op. In this baseline, attention windows are runtime-controlled by the `TrainingManager` object, so args-only patches can be misleading. Once the shrink is applied at the correct control point, the compute reduction shows up as a consistent time-to-target improvement.

**Speed vs margin:** stage3 ws_short shrinking is a real speed lever, but it systematically trades away loss margin (Eval 277’s final loss is ~3.2758, whereas Eval 262 is ~3.2788). This fits the “stability cliff” picture: aggressive compute approximations are viable only when enough margin exists to tolerate variance.

**Robustness knob:** swapping cooldown interpolation for a global LR clamp appears to increase robustness (community replications show cooldown-interp variants can fail on some nodes). The global clamp costs some speed relative to the very fastest lucky pass (Eval 248), but remains sub-93s and easier to reproduce.

**Stage2 seq-len cuts:** mild reductions (e.g., 1920) can pass and further improve time (Eval 276), but more aggressive cuts (1536/1664 here) frequently miss the loss target by ~3e-4–4e-4, indicating a narrow feasible region that is node/seed dependent.

## Related Work

- Archive 1 studied embedding tying/untie schedules under Muon and highlighted stability landscapes but did not address attention-window control points.
- Archive 2 proposed the “Wrapper Synthesis Pattern” and emphasized isolating true algorithmic changes; our work is directly aligned and provides a concrete example of a wrapper-level silent failure mode.
- Archive 3 reported sub-93s configurations via schedule synergies (e.g., untying + LR floor), complementing our focus on window/seq control points.
- Archive 4 introduced boundary-aligned speedups for ImprovedLMHead and explored ws_long cliffs; we extend this by identifying the specific implementation hook required for the window shrink to take effect.
- Archive 5 quantified cross-node variance and formalized the “stability cliff”; our results provide method-level levers (global clamp vs cooldown-interp, ws_short-only shrink) to navigate that cliff.

## Limitation

- The best observed time (Eval 248, 91.92s) is not yet proven robust across nodes; community replications have failed on some workers even with matching high-level settings.
- We do not propose a new optimizer family; improvements come from correct application of existing knobs and schedule shaping.
- Reported standard deviations are computed from three runs (as provided by the Station evaluation); more replications would better characterize tail risk.

## Conclusion

For `baseline_ImprovedLMHead_93`, attention window shrinking must be applied by mutating `TrainingManager.ws_short/ws_long`, not by editing args. Correcting this control point yields a best observed 91.92s (Eval 248). A more robust variant using a global LR floor clamp reaches 92.38s (Eval 262). Controlled ablations show the stage3 ws_short shrink contributes ~0.25s (Eval 262 vs Eval 277) while trading away loss margin, consistent with the Station’s stability-cliff dynamics.

### DIFF
--- a/train_gpt.py
+++ b/train_gpt.py
@@ -1497,8 +1497,9 @@ class TrainingSchedule:
         lr = stage.lr_mul
         cd_start = int(self.scheduled_iterations * (1 - self.cooldown_frac))
         if step >= cd_start:
-            t = min(1.0, (step - cd_start) / (self.scheduled_iterations - cd_start))
-            lr = lr * (1 - t) + 0.1 * t
+            denom = (self.scheduled_iterations - cd_start)
+            t = 1.0 if denom <= 0 else min(1.0, (step - cd_start) / denom)
+            lr = lr * (1 - t) + 0.15 * t
         return lr
 
 # window_sizes are in units of `block_size` tokens (defined in TrainingManager)
@@ -1543,6 +1544,10 @@ class TrainingManager():
     def __init__(self, model):
         self.model = model
         self.block_size = 128
+        self._FULL_SEQ = 2048
+        self._STAGE1_SEQ = 896
+        # ws_* are in block units (block_size=128)
+        self._WS_SHORT_STAGE3 = 2
 
         # - Ordering dictates when to launch reduce/reduce_scatter operations
         # - "sharded" parameters use reduce_scatter/all_gather and "replicated" ones use all_reduce
@@ -1618,6 +1623,14 @@ class TrainingManager():
     def advance_schedule(self, step: int):
         stage, _ = training_schedule.lookup(step)
         self.ws_short, new_ws_long = stage.window_sizes
+        if step == 1:
+            args.train_max_seq_len = self._STAGE1_SEQ
+        elif step == 505:
+            args.train_max_seq_len = self._FULL_SEQ
+
+        # Apply the stage3-only compute cut here.
+        if 1010 <= step < 1515:
+            self.ws_short = self._WS_SHORT_STAGE3
         if new_ws_long != self.ws_long:
             self.model.yarn.apply(self.ws_long * self.block_size, new_ws_long * self.block_size)
             self.model.yarn_paired_head.apply(self.ws_long * self.block_size, new_ws_long * self.block_size)
@@ -1628,6 +1641,10 @@ class TrainingManager():
             self.batch_size = new_batch_size
         else:
             self.train_loader_send_args = None
+
+        # Force loader update at step 1 (batch size unchanged, but seq-len changed).
+        if step == 1:
+            self.train_loader_send_args = (self.batch_size, args.train_max_seq_len, grad_accum_steps)
 
         self.ws_long = new_ws_long
         self.mtp_weights = training_schedule.mtp_weights[step]


varunneal:
Some cool ideas here. Can you also include the updated  train_gpt.py file? Also a few questions:

1. In the 15-run table, what exact settings differ between ImprovedLMHead and Short-Window? (Which of: LR floor+denom guard, Stage1 2048→896→2048, Stage3 ws_short=2, DataLoader refresh are enabled in each column?)
2. Do you have ablations isolating the LR Floor + denom guard and forced DataLoader refresh? I don't see any explanation of what these changes are either -- though I haven't checked through the code in logs yet.

I think human review of the README would be appreciated to ensure that it is parsable for reviewers of this record

dualverse-ai:
> Some cool ideas here. Can you also include the updated train_gpt.py file? Also a few questions:
> 
> 1. In the 15-run table, what exact settings differ between ImprovedLMHead and Short-Window? (Which of: LR floor+denom guard, Stage1 2048→896→2048, Stage3 ws_short=2, DataLoader refresh are enabled in each column?)
> 2. Do you have ablations isolating the LR Floor + denom guard and forced DataLoader refresh? I don't see any explanation of what these changes are either -- though I haven't checked through the code in logs yet.
> 
> I think human review of the README would be appreciated to ensure that it is parsable for reviewers of this record

Thanks for the reminder. We’ve just updated `train_gpt.py`.

To answer your questions:

1. **ImprovedLMHead** corresponds to the baseline SOTA in PR #207, while **Short-Window** (we have changed the name to Eval_248 to avoid confusion) refers to our implementation of the three modifications (we merged four into three). 
2. We've updated all ablation sweeps.

We will also update the README to make everything clearer.

chrisjmccormick:
FYI, in case it's helpful for catching things up--I integrated the PRs prior to yours into your code, and re-ran it for four runs.
The updated code and logs are in a subfolder of my PR #233

ClassicLarry:
This is a fun one! The LLM summary was humorous to read. 

Updating the min_lr from 0.1 to 0.15 is rather elegant. Initially we had a linear decay to 0.1. Later, this was updated to linear decay to 0.1 followed by flat at 0.1 for 40 steps. The new procedure of linear decay to 0.15 and then flat for 40 steps is mathematically equivalent to a linear decay on pace to finish at 0.1, that then halts early once it reaches 0.15. (40 steps is about what the procedure needs to decay 0.05).

Reduces max_seq_len early makes sense. I had proposed this in discussions but at the time didn't feel like the ~0.4s gain was worth its own PR. I've refactored this a bit to align with the design of the repo, and initiating at step 0 instead of step 1.

Updating the 3rd stage short window size to 2 is strange. The short schedule would then follow 1->3->2->6, and the YaRN parameters still update the short RoPE params as if they are increasing (since they use the long window to pull the ratio). The Agent correctly identified that it had generated a lower loss it could then consume with faster execution, but shortening the window in this way is a suboptimal way to achieve that. I reversed this change and have dropped the total step count from 1547 to 1530 and gotten:
losses = [3.2786, 3.2794, 3.2769, 3.2793, 3.2791, 3.2798, 3.2771]
p=0.0091

This gives -1.2s, so 89.8. I will merge this as-is and then follow up with a merge to refactor, drop step count, and drop the ws update. Will give 1.2s credit in readme, because thats the impact of the first two changes.

PR #230, opened 2026-02-12
[New Record] Parallel Residuals (-45 steps, -0.85s)

msisovic:
Adding parallel residuals (for lack of a better name) - attention and MLP read from separate residual streams, but write their output to both streams using learned lambdas.

```python
if block.attn is not None:
  layer_out = block.attn(norm(lane0), attn_args, qkvo_w)
  layer_out = layer_out + skip_bias[i]
  lane0 = resid_lambda[si] * lane0 + layer_out * w_post0[si]
  lane1 = resid_lambda[si] * lane1 + layer_out * w_post1[si]
```

```python
if block.mlp is not None:
  layer_out = block.mlp(norm(lane1), c_fc, c_proj)
  lane0 = resid_lambda[si+1] * lane0 + layer_out * w_post0[si+1]
  lane1 = resid_lambda[si+1] * lane1 + layer_out * w_post1[si+1]
```

This is only applied from block 7 and onwards. This is a tradeoff I made because the the lanes add significant memory bandwidth overhead since we double the amount of lanes we read/write to - while running experiments I noticed the delta between the write strengths between `lane0` and `lane` was significant only in the later layers.

## Inspiration

I converged to this architecture after trying to implement static hyperconnections from the [bytedance paper](https://arxiv.org/abs/2409.19606). However, not only was the overhead huge, but they didn't help convergence at all using the initialization method proposed by the paper (basically starting off equivalent to pre-norm, all lanes symmetrical).

While experimenting, a "round-robin" initialization where I force each layer to pick a favorite lane from the start showed promises, but not that much and the computation/memory overhead was still huge.

I started experimenting which new parameters I could freeze and it turned out that freezing everything except the `H_post` matrix (the one determining how much of the layer output is mixed back into every lane) not only sped steps up, but boosted convergence. Another thing that turned out different from the paper was that 2 lanes was the best in my experiments.

That lead me to simplify the architecture to attn and mlp having separate lanes, which was just an intuition leap since we had 2 lanes anyway.

## Timing and Validation

On an 8xH100 cluster, I re-ran the baseline (based off https://github.com/KellerJordan/modded-nanogpt/pull/216), just to have a representative baseline, along with my changes. Here are the observed results:

```
import scipy.stats
import torch

# First set (Retimed New Record group)
losses1 = [3.2783, 3.2777, 3.2783, 3.2792, 3.2786, 3.2778, 3.2803, 3.2788]
times1 = [91448, 91407, 91518, 91340, 91507, 91542, 91514, 91547]

# Second set (Retimed Current Record group)
losses2 = [3.2807, 3.2789, 3.2789, 3.2778, 3.2783, 3.2774, 3.2801, 3.2783]
times2 = [92394, 92359, 92203, 92367, 92421, 92252, 92389, 92252]

print("p=%.4f" % scipy.stats.ttest_1samp(losses1, 3.28, alternative="less").pvalue)
# p=0.0012
print("losses:", torch.std_mean(torch.tensor(losses1)))
# losses: (tensor(0.0008), tensor(3.2786))
print("time:", torch.std_mean(torch.tensor(times1, dtype=torch.float32)))
# time: (tensor(73.2694), tensor(91477.8750))

print("losses:", torch.std_mean(torch.tensor(losses2)))
# losses: (tensor(0.0011), tensor(3.2788))
print("time:", torch.std_mean(torch.tensor(times2, dtype=torch.float32)))
# time: (tensor(81.3843), tensor(92329.6250))
```

### DIFF
--- a/train_gpt.py
+++ b/train_gpt.py
@@ -1738,6 +1738,7 @@ class TrainingManager():
             "attn_gate_bank": {"optim": "adam",    "comms": "replicated", "adam_betas": [0.9,  0.99]},
             "ve_gate_bank":   {"optim": "adam",    "comms": "replicated", "adam_betas": [0.9,  0.99]},
             "lm_head":        {"optim": "adam",    "comms": "sharded",    "adam_betas": [0.5,  0.95], "wd_mul": 150.},
+            "bigram_embed":   {"optim": "adam",    "comms": "sharded_sparse",    "adam_betas": [0.75, 0.95], "lr_mul": 75.,  "wd_mul": 5.0},
             "w_post":          {"optim": "adam",    "comms": "replicated",    "adam_betas": [0.9,  0.95], "lr_mul": 1.0,  "wd_mul": 0.0},
             "x0_lambda":       {"optim": "adam",    "comms": "replicated",    "adam_betas": [0.9,  0.95], "lr_mul": 1.0,  "wd_mul": 0.0},
             "bigram_lambda":   {"optim": "adam",    "comms": "replicated",    "adam_betas": [0.9,  0.95], "lr_mul": 1.0,  "wd_mul": 0.0},


chrisjmccormick:
Wow! This looks really cool. Double the information capacity of the residual stream, without changing the model dimension. Love that you got this to work!

ClassicLarry:
love this design! Im curious on what values those lambdas learn.

Radical changes like this one opens up the design space for a lot more PRs. 

Holding off on merging until the code is easier to follow. @chrisjmccormick may be looking into that. I think we can push computation into the block.forward() method to make GPT.forward() a bit simpler, and double check all the lambdas don't have extra sizes.

chrisjmccormick:
@ClassicLarry I did indeed! I gave some updates in a comment on my PR, but this is probably the correct place to discuss some of them. 

Here are the cosmetic changes I made:

1. The term `skip_bias`--I think of this quantity as "we're injecting x0 and the bigrams into the residual stream", so I renamed it to `x0_inject`.
2. I removed the excess scalars (I think it was x0_lambdas and bigram_lambdas that were length 22 but only need to be 11).
3. I put back the 's'es in some of the parameters, reverting `x0_lambda` to `x0_lambdas`.
4. I reshaped the `post` weights for clarity. They're now: `[num_layers, 2, 2] = [layer, attn/mlp, lane0/lane1]`
    * I also renamed these to `post_lambdas`, I felt that was more consistent with the codebase.

As far as the location of the residual stream logic (gpt.forward vs. block.forward), this is cosmetic in theory, but had a big impact on speed in my experiments.

The first thing I tried was to define two copies of Block, BlockSingle and BlockParallel.  This looked pretty good.

I think I gained maybe 300-400ms, though, by eliminating block altogether and putting the logic back in gpt.forward.

I think part of why this makes sense is the parameter banks. Our modules are essentially stateless now, and we end up passing in both the inputs and the weights to them, which seems awkward.

I also eliminated the MLP class because it's entirely implemented by the kernel, so I mapped the kernel name to `ReLUSqrdMLP` and we just invoke it directly with
```python
out = ReLUSqrdMLP(norm(lane1), c_fc, c_proj)
```

I did some re-org of the gpt.forward call as well, trying to group some of the lines by task. Mainly, I put all of the "unbind" lines together.

I tried dropping the attention class as well, which I liked because it eliminates the whole `AttnArgs` mechanism. I was having trouble with loss, though, and reverted this in my attempts to fix it. (I think I solved the issue, so we could go back to folding in attention if we want to).

As a counter-argument, it does look pretty clean to just call an attention object. For example, here's what the parallel layers look like currently:

```python
# Parallel: attn reads lane0, mlp reads lane1, both write to both lanes
lane0 = lane0 + x0_inject[i]
out = attn(norm(lane0), attn_args, qkvo_w)
lane0 = resid_lambdas_attn[i] * lane0 + post_lambdas_attn_ln0[i] * out
lane1 = resid_lambdas_attn[i] * lane1 + post_lambdas_attn_ln1[i] * out
post_attn = lane0
out = ReLUSqrdMLP(norm(lane1), c_fc, c_proj)
lane0 = resid_lambdas_mlp[i] * lane0 + post_lambdas_mlp_ln0[i] * out
lane1 = resid_lambdas_mlp[i] * lane1 + post_lambdas_mlp_ln1[i] * out
```

I made a couple interesting behavioral changes as well--I'll put those in another comment here.

chrisjmccormick:
Here are some of the behavioral changes I made.

**Placement of x0/bigram injection**

Larry pointed out that x0/bigram injection now occurs after attention instead of before. 

Injecting prior to attention had a counter-intuitive impact. It seemed to slightly worsen loss (0.0006 --> 12), but was significantly faster (~300ms).
```
This PR (233)            8 89.6050  0.1016    0.0000     NaN  3.2773  0.0009    0.0000 0.0000
Initial refactor         4 89.7460  0.0641    0.1410  0.9917  3.2779  0.0009    0.0006 0.0083  
Inject pre-attn          8 89.4389  0.0980   -0.1661  0.0025  3.2785  0.0012    0.0012 0.0051
```
The impact to loss was small enough that this seemed worth it. Pretty interesting result, though, in terms of interpreting the architecture.

**MLP Output Pre-Multiply**

I noticed that, for the single lane layers, the post-mlp lambda can be pre-multiplied with the output projection (because it's faster to multiply against the weight matrix than the token batch). 

```
ReLUSqrdMLP(norm(lane0), c_fc, post_lambdas_mlp_ln0[i] * c_proj)
```

`c_proj` is pretty large, though (4x larger than the attention output matrix) so the improvement was only ~150ms, and slightly hurt loss. I was having trouble hitting the required confidence, so I chose to drop this change.

> I think this maybe hints at an explanation to the mystery of why @shenberg couldn't get the pre-multiply on attention output to work (without hurting loss). It seems like the numerics must be significant enough that one order of operations produces less rounding error than another, and it may be that it worked in my PR just because of the transposes I did on the weights.

**Attention Output Pre-Multiply**

One of the reasons I like flattened code is that function boundaries can obscure the interactions between components. When you inline attention with the rest of the residual stream logic, it's more obvious that the post attention lambda looks redundant with the sa_lambda. 

I tried pre-multiplying this as well, but had the same problem. This gave a bigger speed improvement than the mlp pre-multiply, but it also hurt loss significantly. I opted instead to drop the scalar altogether, which slightly hurt loss but sped things up enough to justify it, I felt.

**Overall Impact**

I imagine that some of the changes I was making and improvements I was measuring were redundant--coming down to "writing it this way causes the compiler to choose these better kernels". 

The final version I landed on improved over my kernel changes by ~670ms, and increased loss by +0.0009:

```
                      Runs  Steps    Time μ   Time σ  Time +/-   Time p   Loss μ   Loss σ  Loss +/-        p

PR 230 Pll Resids        6   1510   89.7335   0.0781    0.0000      NaN   3.2779   0.0017    0.0000   0.0140
PR 233 Kernels           6   1510   89.3642   0.0860   -0.3693   0.0000   3.2777   0.0007   -0.0002   0.0002
PR 233 Refine Resids     8   1510   88.6966   0.0536   -1.0369   0.0000   3.2786   0.0011    0.0007   0.0032
```

Here's the source code with attention folded into GPT.forward:
[train_gpt-pr233d-flat-forward-regress.py](https://github.com/user-attachments/files/25345904/train_gpt-pr233d-flat-forward-regress.py). It contains the c_proj pre-multiply, so the loss is too high, but if we removed it I think this may work.

Here's the version I timed above, with the c_proj pre-multiply removed, and with attention still its own class:
[train_gpt-pr233f-separate-attn-no-premult.py](https://github.com/user-attachments/files/25345900/train_gpt-pr233f-separate-attn-no-premult.py)

msisovic:
@ClassicLarry Thanks for the feedback! Huge thanks to @chrisjmccormick for the detailed input, help with the refactor and posting all those results!

How do you want to proceed with this? Should I just port over the relevant changes from @chrisjmccormick regarding refactoring? I can isolate the change moving the x0/bigram injection since that was a clear miss in terms of time gain in my PR, and credit should go to Chris. Or we can just ship the changes as part of Chris' PR if it's more convenient @ClassicLarry?

ClassicLarry:
I'd rather include all the refactoring and behavior changes in Chris' PR, but something we are missing is a confirmation at the new step count, since the baseline train.py is now coming in at 1530 steps, and I'm not sure if we can still drop 45 steps off that. Instead of having both you and Chris spend effort to reconfirm that, if just Chris confirms it (or updates to 40 steps or whatever is required) I'm good to merge both and they will be listed as separate records.

msisovic:
BTW, @ClassicLarry you asked about what values the lanes become, here are the logs from a recent run:

```
================================================================================
HYPER-CONNECTION MATRIX ANALYSIS
================================================================================

[resid_lambda] All sublayers: [2.7095187  3.4530241  0.74673814 1.3549066  0.53970873 1.207214
 0.6380573  1.2676467  0.39784363 0.8438695  0.44417647 1.0849047
 1.0488088  0.9060346  0.6652275  1.1177257  0.93591875 1.0355403
 0.83833605 0.9165063  0.6628044  1.4401968 ]
[resid_lambda] Cumulative gain: 0.4053

Layer Type  si      rl     wp0     wp1  |diff|    x0_l    bg_l
-----------------------------------------------------------------
    0 attn   0  2.7095  1.2391     n/a     n/a  0.8541  0.0260
    0  mlp   1  3.4530  0.1498     n/a     n/a  0.8541  0.0260
    1 attn   2  0.7467  1.3962     n/a     n/a  0.8415  0.7608
    1  mlp   3  1.3549  0.3566     n/a     n/a  0.8415  0.7608
    2 attn   4  0.5397  1.3428     n/a     n/a  1.8511 -0.2549
    2  mlp   5  1.2072  0.3796     n/a     n/a  1.8511 -0.2549
    3 attn   6  0.6381  1.3285     n/a     n/a  0.4767  0.1951
    3  mlp   7  1.2676  0.5066     n/a     n/a  0.4767  0.1951
    4 attn   8  0.3978  1.4031     n/a     n/a  1.0460  0.2573
    4  mlp   9  0.8439  0.5629     n/a     n/a  1.0460  0.2573
    5 attn  10  0.4442  1.4483     n/a     n/a  1.8412 -0.2137
    5  mlp  11  1.0849  0.4940     n/a     n/a  1.8412 -0.2137
    6 attn  12  1.0488  1.0000     n/a     n/a  0.0000  0.0500
    6  mlp  13  0.9060  0.5646     n/a     n/a  0.0000  0.0500
    7 attn  14  0.6652  0.8881  2.0019  1.1138  0.0611  0.4224
    7  mlp  15  1.1177  0.4304  0.7258  0.2954  0.0611  0.4224
    8 attn  16  0.9359  0.7477  2.2267  1.4790  0.1164 -0.1782
    8  mlp  17  1.0355  0.1314  0.6794  0.5480  0.1164 -0.1782
    9 attn  18  0.8383  0.7169  2.3687  1.6517 -0.1823 -0.1389
    9  mlp  19  0.9165  0.0390  0.7018  0.6629 -0.1823 -0.1389
   10 attn  20  0.6628  1.4189  2.6797  1.2607 -0.8450  0.4036
   10  mlp  21  1.4402  0.4853  0.4853  0.0000 -0.8450  0.4036
================================================================================
```

Two of the most interesting things to me: the resid lambda's after the first  layer - not necessarily related to my changes, but I haven't checked what the values were before my changes. Second is the differentiation of `wp0` and `wp1` - both favor lane1 (mlp reads from this one), but it's interesting how the attn ones really amplify the signal to more than 2.2x consistently, and how simmilar the learned values are aross the attn layers (~0.7 and ~2.2), which can't fully be said about mlp outputs.

msisovic:
> I'd rather include all the refactoring and behavior changes in Chris' PR, but something we are missing is a confirmation at the new step count, since the baseline train.py is now coming in at 1530 steps, and I'm not sure if we can still drop 45 steps off that. Instead of having both you and Chris spend effort to reconfirm that, if just Chris confirms it (or updates to 40 steps or whatever is required) I'm good to merge both and they will be listed as separate records.

Sound good to me, let's do it like that, thanks

chrisjmccormick:
Yeah, sounds good. 👍

I did a scalar-logging run at one point here:
https://wandb.ai/chrismccormick/Parallel%20Residuals?nw=nwuserchrismccormick

One thing that stood out was the layer 7 residual lambda on attention--it's looking a little chaotic.
image

Also the bigram lambdas in later layers start oscillating:
image

Things may look different, though, after incorporating those new changes to the optimizer settings (like the betas on the LM head).

I've operated under the assumption that we want these scalars to be smooth, but I've become less certain of that. It's possible they help absorb / balance the noise in the weight updates? For example, Karpathy did a bunch of experiments on the best optimizer settings for the x0 and residual lambdas in nanochat, and the residual lambdas look stable in their trajectories but _very_ noisy. 🤷‍♂️

msisovic:
The layer 7 attention resid_lambda being so noisy is pretty interesting - would be interesting to investigate why it's like that. Layer 7 is the first layer where the lanes diverge, but also the lane we take the backout connection out of as well as  coming after layer 6 that has no attention and a skip connection injected into it - would be interesting to see what is the main contributor here, or is it just noise from all these directions adding up.

ClassicLarry:
> BTW, @ClassicLarry you asked about what values the lanes become, here are the logs from a recent run:

Cool, there are some interpretable components here.

The role of attention is to fetch data from other positions. The role of MLP is to process the data that was fetched. EG if attention fetches "Michael" and "Jordan", then MLP can process that into ["Basketball", "Bulls"].

The zigzag pattern in the residual lambdas appears to be a mechanism for the MLP to focus on the unprocessed data from the latest attention module, instead of getting confused with data a prior MLP already processed. But is this data correct? I'd expect layer6 attn resid lambda to sit at 1.1 since its never called.

The wp weights indicate that the parallel residuals is enabling the MLP to build the final prediction, while the attention modules still read from the token context. The transformer doesn't have a good native subtraction mechanism, so to generate its final prediction and overwrite the context scratch pad its build, it produces high activations in the final MLPs to smear over everything its built. However, 'context scratch pad' and 'token prediction' are two different representations and attention prefers to operate on context scratch pad. Layer 8 MLP writes 84% to lane 1, and layer 9 MLP writes 95% to lane 1. I suspect that lane 1 has much higher activations and is dominating the final prediction for most tokens. This way, the very important layer 10 attention can still read from the context scratch pad.

Based on this line of reasoning, it might be favorable to have MLP_9 and Attn_10 both read directly from the residual stream after ATTN_9, and operate in parallel instead of in sequence. There is no real benefit to putting them in sequence if the lambdas are learning over time to ignore 95% of this sequential dependency.

The skip_bias is not applied to layer 6. If this was not directly ablated, I suspect it is suboptimal.

The same residual weight is used for both lanes 0 and 1. This might be limiting the model.

I am suspicious that 1.1 init is still optimal on the residual lambdas. Now that you have twice as many, this pushes up the final activation to 1.1^22 instead of 1.1^11.

msisovic:
@ClassicLarry Actually the init is changed to `sqrt(1.1)` for that exact reason:

```
self.resid_lambda = nn.Parameter(1.1**0.5 * torch.ones(n_sublayers))
```

This also explains your suspicion of the correctness of the converged values, layer 6 attn resid lambda is indeed at it's initial value, as `sqrt(1.1) = 1.0488`

msisovic:
> The skip_bias is not applied to layer 6. If this was not directly ablated, I suspect it is suboptimal.

Yeah, I didn't ablate that, will try and run up some experiments to verify this.

> The same residual weight is used for both lanes 0 and 1. This might be limiting the model.

I played around with this earlier on in my experiments working on this, and it proved to not boost loss while slowing the steps down. The architecture was quite a bit different then so it might be worth re-evaluating.

> Based on this line of reasoning, it might be favorable to have MLP_9 and Attn_10 both read directly from the residual stream after ATTN_9, and operate in parallel instead of in sequence. There is no real benefit to putting them in sequence if the lambdas are learning over time to ignore 95% of this sequential dependency.

This is a very interesting idea to explore as well. I had similar ideas while interpreting the results earlier in experimentation, when it was closer to the HC paper, but didn't solidify it. Good job spotting this one.

I'll come back with some experiment results once I get the time to run these. I'm fine with these being a followup in case you don't want to block on it, but if you want to include it as part of the record for the sake of thoroughness I'm fine with that approach as well.

ClassicLarry:
> I'll come back with some experiment results once I get the time to run these. I'm fine with these being a followup in case you don't want to block on it, but if you want to include it as part of the record for the sake of thoroughness I'm fine with that approach as well.

Follow up works! Then can get another record and it tells the story better of how an architecture evolves over time. People will probably be optimizing this innovation for quite a while.

ClassicLarry:
@msisovic  What do you think of the name 'partitioned hyperconnections'? (hyperconnections where reads are partitioned by stream for attn and mlp)

msisovic:
@ClassicLarry Yeah that works!

ClassicLarry:
Waiting for an 8 H100 node to be available on Prime Intellect so I can merge. Looks like recently there has been a shortage.

ClassicLarry:
Merging at -0.7s, so 89.1s at the -40steps in PR.

YouJiacheng:
btw "round-robin" is similar to AltUp?
https://arxiv.org/abs/2301.13310
image

msisovic:
> btw "round-robin" is similar to AltUp? https://arxiv.org/abs/2301.13310 image

@YouJiacheng It does look very similar, I wasn't aware of this paper, might take a more detailed look later. I'm guessing the update mixing is just slightly different.

PR #235, opened 2026-02-20
[New Record] Crossentropy kernel optimizations (-0.9s)

moof2x:
Kernel optimizations on fused_softcapped_entropy. ~0.9s speedup.

val_loss: [3.2791,3.2778,3.2787,3.2784]
mean: 3.2785
p-value: 0.006

times: [87696ms,87967ms,87672ms,87533ms]
mean: 87717ms

### DIFF
--- a/records/track_1_short/2026-02-23_CrossEntropyKernel/README.md
+++ b/records/track_1_short/2026-02-23_CrossEntropyKernel/README.md
@@ -1,1 +1,8 @@
 Kernel optimizations on fused_softcapped_entropy. ~0.9s speedup.
+
+val_loss: [3.2791,3.2778,3.2787,3.2784]
+mean: 3.2785
+p-value: 0.006
+
+times: [87696ms,87967ms,87672ms,87533ms]
+mean: 87717ms


ClassicLarry:
Could you please add logs? If you didn't check it, I'd recommend trying to drop the final short ws from 6 to 5. Right now its jumping from 4->6 with a YaRN update based on (13/11), with only 40 steps at a low lr to learn the new context len.
edit: also please use PR 223 as the baseline.

moof2x:
> Could you please add logs? If you didn't check it, I'd recommend trying to drop the final short ws from 6 to 5. Right now its jumping from 4->6 with a YaRN update based on (13/11), with only 40 steps at a low lr to learn the new context len. edit: also please use PR 223 as the baseline.

Will do! I'm guessing you mean 233 for the baseline?

ClassicLarry:
> > Could you please add logs? If you didn't check it, I'd recommend trying to drop the final short ws from 6 to 5. Right now its jumping from 4->6 with a YaRN update based on (13/11), with only 40 steps at a low lr to learn the new context len. edit: also please use PR 223 as the baseline.
> 
> Will do! I'm guessing you mean 233 for the baseline?

yea, 233. My experience with micro-tweaks like 0.17 min_lr and 0.58 cooldown_frac is searching for 0.01p value changing a bunch of stuff, and then when good runs come along you keep whatever microtweaks were active at the time. But sometimes those microtweaks don't hold up on fresh ablations. Curious if these are still meaningfully helpful on the new baseline.

moof2x:
I ran 10 runs on the current master and got these losses:

losses [3.2808, 3.2799, 3.2785, 3.2809, 3.2787 ,3.2806, 3.2811, 3.2791, 3.2773, 3.2805]
mean 3.27974
p-value 0.2676

Maybe PR #230 and #233 didn't play nicely together? Or maybe something on my side. Looking at the PRs they both had a good safety margin so getting p < 0.01 wasn't hard. It seems like that margin has disappeared.

ClassicLarry:
I think we are probably sitting on zero margin right now and the runs for 233 were a bit lucky. Which is fine but it means that future PRs have a slightly higher bar to clear to get merged. May need 0.5-1s improvement to reliably get 0.01 p. The ML change in 233 was to remove the lambdas after attn on the single lanes, which could be revisited or slightly tweaked. Unfortunately that does mean that this PR might need a bit more stacked on top of it to clear the threshold.

moof2x:
I am determined to get a record! I have some kernel optimizations that should get this close to a second in speedup. Logs pending....

moof2x:
I reverted the learning rate/window size/sequence length changes. With the updated baseline they were still getting about 0.25s of speedup, but the was a tiny increase in loss so I couldn't get significance with the margin being so slow now.

Added in my kernel optimizations instead, and got some good results. About a second of improvement! Also added logs.

ClassicLarry:
neat! will run later.

ClassicLarry:
Merging at -0.9s for 87.2. I'm not worried about loss here since its a kernel update, but I think that the 4 tests in main underestimate the true loss a bit. The 10 runs you did earlier seem more representative. 

Later I'll push some improvements that came out of the partitioned hyperconnections discussion that should bring the loss back more cleanly under 3.28.

PR #237, opened 2026-02-25
fix: replace bare except with except Exception in SOAP eigendecomposition

haosenwang1018:
Replace bare `except:` with `except Exception:` in the SOAP optimizer's eigendecomposition fallback. The float64 fallback for `torch.linalg.eigh` should only catch numerical errors, not `KeyboardInterrupt`.


ClassicLarry:
Not looking to merge in syntax changes in arbitrary old logs.

PR #246, opened 2026-03-22
New Record: Reduce max_num_docs, -500ms

chrisjmccormick:
# Reduce Varlen max_num_docs

While working on refactoring `nanochat` for FlashAttention varlen ([here](https://github.com/karpathy/nanochat/pull/663)), I discovered that the size of the `cu_seqlens` tensor (which contains the document offsets in our 1D training batch) matters more than expected. 

Analyzing the fineweb dataset and better tailoring the size of `cu_seqlens` gave a ~500ms speed improvement.

Additionally, I tried some recent `nanochat` changes and found that the reduction of Muon's second beta from `0.95 --> 0.9` slightly improved loss, so I included that as well (no reduction in steps).

```
                    Runs  Steps   Time μ   Time σ  Time +/-   Time p   Loss μ   Loss σ  Loss +/-        p
baseline               8   1490  86.7400   0.0497    0.0000       NaN  3.2787   0.0015    0.0000   0.0218 
max_num_docs           4   1490  86.2422   0.0479   -0.4977   0.0000   3.2785   0.0012   -0.0002   0.0444 
muon-beta2             4   1490  86.7758   0.0274    0.0358   0.9296   3.2785   0.0007   -0.0002   0.0122 
combined (this PR)     8   1490  86.2246   0.0300   -0.5154   0.0000   3.2784   0.0013   -0.0003   0.0054 

baseline:
  losses = [3.2781, 3.2800, 3.2786, 3.2773, 3.2811, 3.2773, 3.2800, 3.2772]
  times  = [86.8210, 86.7020, 86.6940, 86.7430, 86.7960, 86.6970, 86.7010, 86.7660]

This PR:
  losses = [3.2788, 3.2785, 3.2792, 3.2762, 3.2792, 3.2795, 3.2794, 3.2766]
  times  = [86.2030, 86.2360, 86.2820, 86.2020, 86.1960, 86.2340, 86.2000, 86.2440]
```

### Choosing max_num_docs

We need `cu_seqlens` to be a fixed size for `torch.compile`. However, because we have a separate compute graph per batch size, we can safely choose a different `cu_seqlens` size for each batch size.

To determine appropriate sizes, I analyzed the first 10 shards (1.45M documents).

For token counts overall:
* median 403
* mean 690 
* min 32

I constructed the actual batches used and counted the number of documents in each, across our three batch sizes / stages:

| Stage | tokens/rank | current `max_num_docs` | observed max | proposed | savings |
|-------|-------------|----------------------|--------------|----------|---------|
| Stage 1 | 16,384 | 128 | 56 | **64** | -64 slots (50%) |
| Stage 2 | 32,768 | 128 | 90 | **96** | -32 slots (25%) |
| Stage 3 | 49,152 | 256 | 121 | **128** | -128 slots (50%) |

I've hardcoded those batch size --> num docs mappings, but with a fallback to the original method of rounding up conservatively if a different batch size is encountered.

### DIFF
--- a/records/track_1_short/2026-03-22_VarlenMaxDocs/README.md
+++ b/records/track_1_short/2026-03-22_VarlenMaxDocs/README.md
@@ -1,0 +1,36 @@
+# Reduce Varlen max_num_docs
+
+While working on refactoring `nanochat` for FlashAttention varlen, I discovered that the size of the `cu_seqlens` tensor (which contains the document offsets in our 1D training batch) matters more than expected. 
+
+Analyzing the fineweb dataset and better tailoring the size of `cu_seqlens` gave a ~500ms speed improvement.
+
+Additionally, I tried some recent `nanochat` changes and found that the reduction of Muon's second beta from `0.95 --> 0.9` slightly improved loss, so I included that as well (no reduction in steps).
+
+```
+                    Runs  Steps   Time μ   Time σ  Time +/-   Time p   Loss μ   Loss σ  Loss +/-        p
+baseline               8   1490  86.7400   0.0497    0.0000       NaN  3.2787   0.0015    0.0000   0.0218 
+max_num_docs           4   1490  86.2422   0.0479   -0.4977   0.0000   3.2785   0.0012   -0.0002   0.0444 
+muon-beta2             4   1490  86.7758   0.0274    0.0358   0.9296   3.2785   0.0007   -0.0002   0.0122 
+combined (this PR)     8   1490  86.2246   0.0300   -0.5154   0.0000   3.2784   0.0013   -0.0003   0.0054 
+```
+
+### Choosing max_num_docs
+
+We need `cu_seqlens` to be a fixed size for `torch.compile`. However, because we have a separate compute graph per batch size, we can safely choose a different `cu_seqlens` size for each batch size.
+
+To determine appropriate sizes, I analyzed the first 10 shards (1.45M documents).
+
+For token counts overall:
+* median 403
+* mean 690 
+* min 32
+
+I constructed the actual batches used and counted the number of documents in each, across our three batch sizes / stages:
+
+| Stage | tokens/rank | current `max_num_docs` | observed max | proposed | savings |
+|-------|-------------|----------------------|--------------|----------|---------|
+| Stage 1 | 16,384 | 128 | 56 | **64** | -64 slots (50%) |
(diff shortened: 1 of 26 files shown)


varunneal:
Awesome. At some point I wired up FA3 to allow max_num_docs to be a tensor (so it would be varying) but I didn't notice a big time advantage. I might go for the refactor again and see if I can raise a PR on the official repo

chrisjmccormick:
> Awesome. At some point I wired up FA3 to allow max_num_docs to be a tensor (so it would be varying) but I didn't notice a big time advantage. I might go for the refactor again and see if I can raise a PR on the official repo

Oh, cool! So, keep the size of `cu_seqlens` fixed, but additionally specify the true number of documents in the batch, so that FA could be smart and avoid whatever overhead it is that we're solving for here?

ClassicLarry:
Cool nice find! Im going to merge this at 86.1-0.5 for 85.6s without running a second validation, as there's an H100 shortage and you already have a bunch of clean merges.

varunneal:
how did the speed of the varlen function compare to the speed of the fixed-seq size function in nanochat (eg for a similar batch size)?

chrisjmccormick:
@ClassicLarry Thanks Larry! Yeah, I think I had to wait something like two weeks to get an 8xH100 instance to run both this and the nanochat PR. I wonder if parameter-golf is to blame--the amount of activity on that challenge is psychotic. There are currently about 900 PRs. 😂

chrisjmccormick:
> how did the speed of the varlen function compare to the speed of the fixed-seq size function in nanochat (eg for a similar batch size)?

@varunneal In nanochat, it sped up the training run by 1.8% overall. nanochat uses a fixed micro-batch size of 32K tokens, corresponding to our 2nd stage. 

I grabbed trace files to look at the improvement, but it was a bit of a messy picture. Maybe partly because it uses fp8 training, which adds small kernels everywhere. 

But also varlen's compute requirements presumably depend on the actual sequence lengths of the documents in the batch, and even though the traces (with and without varlen) were on the same training steps, the dataloaders behave differently (baseline nanochat does a best-fit strategy and chops more tokens), so it's not a perfect comparison. 

Best I could tell--I think the savings are primarily on the long context attention layers.

chrisjmccormick:
And just to throw this out there--I put together a blog post on how varlen works as well:

http://mccormickml.com/2026/03/23/flashattention-varlen/

PR #250, opened 2026-04-03
New Record: remove unneeded float casts

alexandermorgan:
This branch runs ~2.5s faster than my runs of the unchanged master branch. The main thing it does is remove some unnecessary float casts in each step of training, specifically lines 1273-4:
```
ag = [w.bfloat16() for w in self.attn_gate_bank.unbind(0)]
veg = [w.bfloat16() for w in self.ve_gate_bank.unbind(0)]
```
These two `.unbind(0)` objects are already iterables of floats, so this is a bunch of casting calls that are effectively no-ops, and also unneeded unpacking into new lists. Those immediately get repacked again anyway, so this pr just skips the intermediate lists:
```
ag = self.attn_gate_bank.unbind(0)
veg = self.ve_gate_bank.unbind(0)
```

On top of this primary change, this pr also makes a lot of little python improvements. For example, all the dataclass classes get instantiated with `slots=True` to save 1 dictionary per class instance.

This line is also raised to be outside the for loop so it doesn't get recalculated unnecessarily:
`self._eff_wd_t.fill_(p_cfg.wd_mul * p_cfg.weight_decay * p_cfg.lr)`

dict.keys() objects are either used more idiomatically or omitted when not necessary (a dictionary is already a key iterator). And the function `next_multiple_of_n` is refactored to calculate directly rather than using a generator that has to pass over a bunch of wrong answers first.

Using the `+` operator with more than 2 lists in python is a bad idea so the way several lists are poured is refactored, e.g. this creates 5 total lists:
`ends = [0] + [round(c * scheduled_iterations) for c in accumulate(s.duration for s in stages[:-1])] + [self.total_steps]`
so I refactored it to this which is just 2 lists:
`ends = [0, *[round(c * scheduled_iterations) for c in accumulate(s.duration for s in stages[:-1])], self.total_steps]`

## Timings

I ran the unchanged master branch and this pr branch (remove_float_casts) 3 times on two different days. My timings were:
master -- 88185ms, 88646ms, 88579ms
remove_float_casts -- 85640ms, 86089ms, 86053ms
which is an average of 2.543s faster with remove_float_casts.


ClassicLarry:
Cool, will retime later. If I can confirm the >2s gain that will be awesome. and kinda wild, since the op changes are mostly touch-ups at first glance.

alexandermorgan:
Yeah I agree, the impact of this one is larger than I anticipated.

ClassicLarry:
Unfortunately I haven't been able to replicate a speedup on this one yet.

Rerunning master:
step:1490/1490 val_loss:3.2786 train_time:86126ms step_avg:57.80ms
step:1490/1490 val_loss:3.2778 train_time:86132ms step_avg:57.81ms

This PR: 
step:1490/1490 val_loss:3.2778 train_time:86145ms step_avg:57.82ms
step:1490/1490 val_loss:3.2797 train_time:86112ms step_avg:57.79ms
step:1490/1490 val_loss:3.2771 train_time:86088ms step_avg:57.78ms

How I typically run these is spin up ubuntu machine 8H100 SXM on prime intellect,
setup pip and install jupyterlab, then run 
```
git clone https://github.com/KellerJordan/modded-nanogpt.git && cd modded-nanogpt
# downloads only the first 900M training tokens to save time
./run.sh
```

My best guess is that the torch version or hardware you are testing on is giving a poor compilation on the baseline for some reason. I wouldn't expect the particular changes here to be boosting by >2s, unless there was some weird compilation artifact going on.

ClassicLarry:
I am happy to merge this in as a cleanup PR, but it wouldn't be added to the leaderboard until I or another contributor can confirm the speedup.

alexandermorgan:
My apologies for the false positive! Yes, please merge this as a cleanup PR. Perhaps these changes will also make the timings more predictable for other speed-runners trying things out.
I'm curious, in your setup why do you need to install jupyterlab manually?

PR #253, opened 2026-04-08
New record: Muon paired-head groups (-10 steps, -.4s)

samacqua:
# Paired-head Muon groups

Treat each paired-head in K and V as its own Muon group, reducing loss/step and allowing us to reduce step count by 10 and still meet 3.28 loss target.

```bash
  baseline-1490     4   1490  93.0205   0.1629    0.0000      nan   3.2782   0.0010   -0.0018    0.0187
  baseline-1480     8   1480  92.5491   0.0588   -0.4714   0.0040   3.2800   0.0017    0.0000    0.5000
        this PR     8   1480  92.6259   0.2129   -0.3946   0.0038   3.2788   0.0009   -0.0012    0.0029
```

Previously, Muon treated each of Q, K, V, O as one `(hdim, hdim)` group per layer, lumping all heads together within each group. Given heads often specialize, it seemed intuitive that treating each *head* as its own Muon group would further improve optimization. I split K/V into per-**head-pair** groups (because of paired head attention in some layers). This gives each head pair its own independent polar decomposition. Treating each head-pair in K and V as its own Muon group improved loss per step while increasing time/step by ~.1% (this overhead can probably be reduced, splitting Muon groups further should reduce flops).

Now, we can reach the target loss in 10 less steps, saving ~0.4s. If we reduce the baseline by 10 steps without adding the per-head-pair Muon groups, the loss does not meet the target.

**Note on timing:** not sure why my 8xh100 is so slow -- ~8% slower per step for the baseline. However, given that the speedup comes from reducing steps, I think this will hold in a more controlled setup. Also, I see there have been other optimizations that have not yet been merged, but they are orthogonal to this PR.

### DIFF
--- a/records/track_1_short/2026-04-08_PairedHeadMuon/README.md
+++ b/records/track_1_short/2026-04-08_PairedHeadMuon/README.md
@@ -1,2 +1,17 @@
-# Paired head muon
+# Paired head Muon groups
 
+Treat each paired-head in K and V as its own Muon group, reducing loss/step and allowing us to reduce step count by 10 and still meet 3.28 loss target.
+
+
+```bash
+                 Runs  Steps   Time μ   Time σ  Time +/-   Time p   Loss μ   Loss σ  Loss +/-   Loss p
+  baseline-1490     4   1490  93.0205   0.1629    0.0000      nan   3.2782   0.0010   -0.0018    0.0187
+  baseline-1480     8   1480  92.5491   0.0588   -0.4714   0.0040   3.2800   0.0017    0.0000    0.5000
+        this PR     8   1480  92.6259   0.2129   -0.3946   0.0038   3.2788   0.0009   -0.0012    0.0029
+```
+
+Previously, Muon treated each of Q, K, V, O as one `(hdim, hdim)` group per layer, lumping all heads together within each group. Given heads often specialize, it seemed intuitive that treating each *head* as its own Muon group would further improve optimization. I split K/V into per-**head-pair** groups (because of paired head attention in some layers). This gives each head pair its own independent polar decomposition. Treating each head-pair in K and V as its own Muon group improved loss per step while increasing time/step by ~.1% (this overhead can probably be reduced, splitting Muon groups further should reduce flops).
+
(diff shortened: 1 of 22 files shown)


ClassicLarry:
Cool, will merge later at -0.3s based on retiming. Since last record was 84.7 that will put this at 84.4s.

Rerunning on the H100s I happened to get today:
```
import scipy.stats
import torch
losses = [3.2798, 3.28, 3.2793, 3.2801, 3.2805, 3.2791, 3.2768, 3.2789, 3.279, 3.2801, 3.2787, 3.2781]

times = [84.858, 84.802, 84.883, 84.885, 84.911, 84.911, 84.948, 84.896, 84.907, 84.968, 84.885, 84.899]

# p=0.0105
# losses: (tensor(0.0010), tensor(3.2792))
# time: (tensor(0.0417), tensor(84.8961))
```

This one is close to the border of what would constitute a new record, but I think it may open up a small opportunity for faster comms with bundling some of the qk/vo params, and also requiring less data for training is nice.

ClassicLarry:
Also to add- many people, including myself, have tried the per-head orthonormalization, as it seemed more intuitive at first than bundling all heads. However all our attempts failed, including ones that broke out Q and K etc. So, nice execution on finding a combination that worked out.

ClassicLarry:
I had added some ramblings on this topic of the attention heads here: https://github.com/KellerJordan/modded-nanogpt/discussions/23#discussioncomment-15233803. Your paired head variant is perhaps better than what everyone was trying, in that it makes the muon op closer to square.

ClassicLarry:
I noticed the readme says K and V but the code does it for K and Q. I tested a version where qkv is a single parameter bank and the paired muon gets applied over it, which never needs to get concatenated to o_w. results were about the same.

samacqua:
> I noticed the readme says K and V but the code does it for K and Q

Yes, typo -- thanks.

One thing to add to this PR in case anyone is looking back at this: I listed paired-head attention as a motivation and explanation for why using paired-head muon groups work but the single-head groups do not. However, this isn't correct. I ran the ablation of using different pairs than in paired head attention (i.e. paired-head attention pairs heads 0&1, 2&3, 4&5 while muon groups 0&3, 1&4, 2&5) and the loss was actually a bit better.

PR #254, opened 2026-04-09
New Record: Polar Express Optimization (~440ms)

DeepReinforce:
All changes are confined to the **Polar Express orthogonalization** step and the **sparse communication** routines used during distributed training. The model architecture, hyperparameters, training schedule, and all other code remain identical.

### 1. Gram-Trace Normalization with Diagonal Fold

**Before:** The spectral norm of the input matrix is bounded by dividing by the Frobenius norm — a full-matrix operation that materializes a normalized copy every iteration.

```python
X = X / (X.norm(dim=(-2, -1), keepdim=True) * (1 + 2e-2) + 1e-6)
```

**After:** Normalization uses the Gram matrix diagonal (trace of X^T X), which is already computed for the iteration. The scale factor is folded into the first polynomial coefficient via diagonal addition, avoiding a separate full-matrix division.

```python
scale = A.diagonal(dim1=-2, dim2=-1).sum(dim=-1, keepdim=True, dtype=torch.float32)
scale.sqrt_().mul_(_POLAR_SAFETY_SCALE).add_(_POLAR_EPS).reciprocal_()
A.mul_((scale * scale)[..., None].to(dtype=A.dtype))
# ... first iteration ...
B.diagonal(dim1=-2, dim2=-1).add_(_P0A)  # fold aI into polynomial
B.mul_(scale[..., None].to(dtype=B.dtype))
```

This eliminates the `aX + X @ B` fused addmm/baddbmm pattern (which required referencing X twice and triggered defensive copies in PyTorch) and replaces it with a single `bmm` per iteration.

### 2. Specialized Tall-Matrix Paths (MLP Weights)

The MLP weight bank has shape (3072, 768) per matrix — a 4:1 tall aspect ratio. The baseline uses the same iteration loop for all shapes, performing 5 tall-matrix multiplications per step.

**After:** Two dedicated tall-matrix paths are introduced:

- **3-step accumulated transform** (`_polar_express_tall_transform_batched`, used for steps < 500): Propagates the Gram and accumulated transform entirely on the small n×n factor, touching the tall m×n matrix only twice (initial XTX, final bmm). This cuts 3 expensive tall-matrix bmm operations per step.

- **5-step standard path** (`_polar_express_tall_standard_batched`, used for steps ≥ 500): Uses the full 5-step self-correcting iteration with Gram-trace normalization for higher quality convergence in later training.

The transition at step 500 (`_POLAR_SWITCH_STEP`) balances speed in early training against quality in later training.

### 3. Fully Unrolled Iterations

**Before:** Polar Express iterations are written as a Python `for` loop over the coefficient list, which prevents `torch.compile` from fully optimizing memory reuse across iterations.

**After:** Each specialized function has all 5 (or 3) iterations fully unrolled with pre-extracted scalar coefficients (`_P0A`, `_P0B`, ..., `_P4C`). Combined with `fullgraph=True`, this allows the compiler to fuse operations across iteration boundaries and optimize buffer reuse.

### 4. Sparse Communication Improvements

Several micro-optimizations to the distributed sparse gradient communication:

- **Cached partition boundaries** (`_sparse_partition_boundaries`): The `searchsorted` boundary array is computed once per (N, world_size) pair and reused, avoiding repeated allocation.
- **In-place index slicing**: Local rank's rows are excluded via `narrow` + conditional `copy_` instead of `torch.cat` with two slices, reducing CPU-side allocations.
- **Local gradient merge**: `sparse_comms_merge_gradients` now operates on a `narrow`ed local slice with adjusted indices (`recv_idx.sub(start)`), avoiding the final full-tensor slice.

## Results

All experiments run on 8×H100 GPUs with identical seeds, hyperparameters, and training schedule. Each configuration was run multiple times to assess variance.

| Configuration | Runs | Mean Time (ms) | Mean Val Loss | Time Saved |
|---|---|---|---|---|
| **Baseline (03/22/26)** | 5 | 85,658 | 3.2786 | — |
| **Baseline + Polar Express Optimization** | 5 | 85,219 | 3.2777 | **439 ms (0.51%)** |
| **PR#251** | 4 | 84,930 | 3.2788 | — |
| **PR#251 + Polar Express Optimization** | 5 | 84,545 | 3.2795 | **385 ms (0.45%)** |

All configurations achieve val_loss < 3.28.

### DIFF
--- a/records/track_1_short/2026-04-09_Polar_Express_Optimization/README.md
+++ b/records/track_1_short/2026-04-09_Polar_Express_Optimization/README.md
@@ -1,0 +1,84 @@
+# Polar Express Optimization
+
+## Optimizations
+
+All changes are confined to the **Polar Express orthogonalization** step and the **sparse communication** routines used during distributed training. The model architecture, hyperparameters, training schedule, and all other code remain identical.
+
+### 1. Gram-Trace Normalization with Diagonal Fold
+
+**Before:** The spectral norm of the input matrix is bounded by dividing by the Frobenius norm — a full-matrix operation that materializes a normalized copy every iteration.
+
+```python
+X = X / (X.norm(dim=(-2, -1), keepdim=True) * (1 + 2e-2) + 1e-6)
+```
+
+**After:** Normalization uses the Gram matrix diagonal (trace of X^T X), which is already computed for the iteration. The scale factor is folded into the first polynomial coefficient via diagonal addition, avoiding a separate full-matrix division.
+
+```python
+scale = A.diagonal(dim1=-2, dim2=-1).sum(dim=-1, keepdim=True, dtype=torch.float32)
+scale.sqrt_().mul_(_POLAR_SAFETY_SCALE).add_(_POLAR_EPS).reciprocal_()
+A.mul_((scale * scale)[..., None].to(dtype=A.dtype))
+# ... first iteration ...
+B.diagonal(dim1=-2, dim2=-1).add_(_P0A)  # fold aI into polynomial
+B.mul_(scale[..., None].to(dtype=B.dtype))
+```
+
+This eliminates the `aX + X @ B` fused addmm/baddbmm pattern (which required referencing X twice and triggered defensive copies in PyTorch) and replaces it with a single `bmm` per iteration.
+
+### 2. Specialized Tall-Matrix Paths (MLP Weights)
+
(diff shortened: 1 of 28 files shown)


ClassicLarry:
The most recently merged in PR changed the shape of the matrices for Q and K (mentioning in case this impacts this PR). I will look at this one in more detail later, looks like some neat stuff.

ClassicLarry:
Reading this more closely. Since it changes the learning algorithm, it needs to show p value <0.01 to be merged.

PR #255, opened 2026-04-15
Implemented recently proposed Exclusive Self Attention in train_gpt.py

Mithun-033:
Implements XSA by removing the component of the attention output aligned with the token’s own value vector, encouraging attention to focus on contextual information rather than self-copying. This is a 2 line addition to the Causal Attention block .


kroggen:
have you paid attention that all the code around uses spaces between variables and operators?

does it work with value embeddings? (this repo uses)

have you run the benchmark? do you have results?

PR #257, opened 2026-04-20
Fix record formatting

MarktHart:
It seems self explanatory, but to be precise: changed `-4` to `-04` to fix date ordering.


ClassicLarry:
thanks!

PR #259, opened 2026-04-22
New record: MUDDFormer connections (-65 steps, -2.58s)

Lisennlp:
# MUDDFormer connections

Add MUDD (Multiway Dynamic Dense Connections, [Xiao et al. 2025](https://arxiv.org/abs/2502.12170)) to the last two layers, letting us cut 50 steps off the schedule while still hitting the 3.28 loss target.

```bash
  baseline-1480    10   1480  84.7039   0.0449    0.0000      nan   3.2788   0.0011    0.0000    0.0036
  baseline-1430    10   1430  81.6762   0.0611   -3.0277   0.0000   3.2867   0.0014   +0.0080    1.0000
        this PR    10   1430  83.2020   0.0326   -1.5019   0.0000   3.2783   0.0016   -0.0004    0.0040
```

MUDD adds input-dependent dense connections between transformer blocks: for each target layer i, an MLP on the current hidden produces per-position weights over a set of earlier hiddens, and the weighted sum is injected back into the residual / attention inputs. The original paper routes a separate dynamic-weighted mixture into each of Q, K, V and the residual at every layer, which is too expensive at this model scale. This PR uses a trimmed variant that keeps most of the quality gain at much lower cost:

Only the last two layers (i ∈ {N-2, N-1}) consume MUDD signals. Earlier blocks are unchanged, so the bulk of the forward/backward is untouched.

Small source-hidden subset per layer. Layer N-2 mixes over {x0, h7, h9} (3 sources, where x0 is the pre-layer-0 hidden and h9 is the current layer's post-MLP output — a self-reference); layer N-1 only mixes over {x0, h9}. A much smaller subset than the fully-dense MUDD in the paper, which avoids quadratic growth in mixture cost and keeps most of the quality gain at this model scale.

Two channels instead of four. We drop the Q/K MUDD adds from the paper and keep only v_mudd (added into V) and residual_mudd (added into the residual stream). The Q/K channels empirically contributed little while costing two extra routed mixtures per layer.

V-fusion, no extra projection. v_mudd is produced in the same (B,T,H,D_head) shape as V and added after the existing QKV matmul, so MUDD does not introduce a new per-channel linear into attention.

Dense params on Adam, low LR. dense_w1 / dense_w2 / dense_bs are replicated-Adam with betas=(0.9, 0.99) and lr_mul=0.25 (no WD on the biases). dense_w2 and dense_bs are zero-initialized, so at init the MUDD contribution (both v_mudd and the residual delta) is exactly zero — MUDD starts as a no-op and has to learn its way in.

Effect at matched steps: adding these connections at step 1430 recovers (slightly beats) the step-1480 baseline loss, while baseline-1430 without MUDD is clearly above target (Δloss +0.0080, p≈1.0). The MUDD forward/backward adds ~1.5s vs baseline-1430 (~1.05 ms/step), but since we now need 50 fewer steps the net wall-clock is -1.5s vs baseline-1480 at equal-or-better loss.

Things I didn't try / would expect to squeeze more
- Putting MUDD on more than 2 layers (maybe just 1 more block mid-network). Didn't want to inflate the overhead budget.
- Tuning lr_mul / betas for the dense params — I took a reasonable guess and stopped.
- Re-enabling a Q or K channel but only on one layer — might be worth it now that the Q/K projections are already fast.
- dense_w1 is kaiming-init; wonder if small-init would learn faster given w2 starts at zero anyway.
Timing was on 8xH100.

(Disclosure — I'm one of the MUDDFormer authors, but this is a pretty aggressive cut-down of the original recipe to fit the speedrun. Very curious to hear what people think, and would love suggestions on where to squeeze more.)


ClassicLarry:
wow. reading paper in detail!

ClassicLarry:
Super cool! lot to think about here.

Conceptually I am following the implementation here to be:

before the lm_head we previously had 
`x = x - backout_lambda * x_backout`
Which was equivalent to 
`x = x - backout_lambda * x_7`

Now before the lm_head we have 
`x = x - backout_lambda * x_7 + m9 * x_9 + m0 * x_0`

Then we have another skip connection point on layer 9 (with new m coefficients):
`x = x + m9 * x + m0 * x_0 + m7 * x_7`

And we also have a skip connection to our last layer attention values:
`v = v + (mv0 * x_0 + mv7 * x_7 + mv9 * x_9)`

In the context of this speedrun, the novelty is:
1. Extending the web of existing residual skip connections
2. Adding a skip connection from residuals to the final value embedding. (which breaks some of the symmetry the cached activation introduced)
3. Introducing a more powerful operator for skip connection magnitude. The model currently has lambdas, and data dependent lambdas via sigmoid(X dot w). the M's in the equations above use an MLP of dim (64?) with a bias. So each measures 64 directions in the activation, whereas the current sigmoid approach only measures 1. It is also compute efficient because it shares the input projection.

Based on this understanding, some natural extensions:
1. We have a redundant addition with x = x + m9 * x, can do (1+m9) * x
2. unify the equation before the lm_head to use m7 instead of backout_lambda.
3. It seems surprising to add a linear combination of residuals directly to an attention value since these live in different subspaces. I'm surprised this works-maybe I'm missing a linear transform thats already there? Or maybe worth adding one to convert between subspaces. Or maybe I need to rethink my understanding haha.
4. Use the same mudd coef approach on the value embed gates, instead of sigmoid(dot).
5. Use the same mudd coef approach on the 3->6 skip connection.

I am finding the generic code style hard to follow. Would like to update the code to make the sparse connectivity pattern explicit. I can take a stab at this later.

Lisennlp:
Thanks for the detailed read — your decomposition is essentially right. Two small clarifications:

- The `m9` in the "before lm_head" line and the `m9` in the "layer 9 skip" line are independent params: the former is layer 10's MUDD coef over `x_9`, the latter is layer 9's coef over `x_9` (self).
- The 64-dim figure is incidental: `initer_dim = ((max_C * L) // 64 + 1) * 64`, which rounds up to 64 here only because `(C, L) = (2, 3)` is tiny. If either grew it would step up to 128 etc.

Otherwise the three-point framing of the novelty — wider skip web, residual→V skip, and 64-dim multi-direction gate vs the 1-D `sigmoid(x·w)` — is exactly how I think about it. Reframing MUDD as "a more expressive form of the existing skip primitive" is honestly cleaner than the way I wrote it up.

Going through the extensions:

1. **`(1 + m9) * x` fuse — agreed.** The self-reference falls out of the generic "loop over L sources" path; folding it is an easy win.
2. **Unify `backout_lambda` into `m7` before lm_head — tried it on your suggestion.** Both loss and wall-clock came out slightly better than the current setup, but the gap is small enough that I'd want more seeds before claiming it's real. Happy to fold it in either way if you'd like — even at face value it's a cleaner formulation, and the trend is at least in the right direction.

3. **V add living in a different subspace — really good question, and I was partly winging it here.** You're right that there's no projection: `v_mudd` is a `(B,T,D)` weighted sum of residuals, then a bare `view → (B,T,H,D_head)`. The reasons I think it survives, roughly in order of confidence:
   - V is itself `W_V @ x`, so V's subspace is *defined* relative to the residual; adding a learned mixture of residuals back into V is less foreign than it looks.
   - `dense_w2` / `dense_bs` are zero-init, so MUDD starts as a no-op — the model can choose 0 if the V add is harmful.
   - Coefficients are per-token, so the signal can be routed sparsely.

   The paper's formulation does have a small linear inside the dynamic aggregation before adding to V; dropping it was a wall-clock decision, not a principled one. A tiny per-head re-mix is genuinely worth testing — might recover loss for less wall-clock than the steps it would save.
4. **MUDD coef on the VE gates and the 3→6 skip — both good targets.** VE gates in particular are the most direct fit since they're already per-token gates. I scoped this PR to "two layers + V add" deliberately to keep the diff small and the wall-clock honest.
5. **Refactor for explicit sparse connectivity — please do, would love that.** The current generic loop was written so I could sweep subsets quickly; now that the connectivity is frozen at `{x0, h7, h9}` / `{x0, h9}`, an explicit-edges version would be much clearer and likely trim a little overhead too.

ClassicLarry:
Ok, I'll update the syntax later! Any updates to the logic can be treated as a separate PR and new record.

Lisennlp:
Posted v2 (new commit + records under `records/track_1_short/2026-04-22_MuddFormer/this_pr_v2/`). README has the full table; headline:

```bash
  baseline-1480    10   1480  84.7039   0.0449    0.0000      nan   3.2788   0.0011    0.0000    0.0036
  baseline-1430    10   1430  81.6762   0.0611   -3.0277   0.0000   3.2867   0.0014   +0.0080    1.0000
     this_pr_v1    10   1430  83.2020   0.0326   -1.5019   0.0000   3.2783   0.0016   -0.0004    0.0040
this_pr_v1-1415    10   1415  82.1446   0.0342   -2.5593   0.0000   3.2815   0.0012   +0.0027    0.9982
     this_pr_v2    10   1415  82.1289   0.0386   -2.5750   0.0000   3.2781   0.0010   -0.0007    0.0001
```
So v2 is **-2.58s vs baseline-1480** at strictly-better loss, and **-1.07s / -15 steps vs v1** at slightly-better loss. (`Loss p` = `scipy.stats.ttest_1samp(losses, 3.28, alternative="less").pvalue`, 10 seeds each.)

What's in v2, mapped to your suggestions:

1. **`(1 + m_r9) * x` self-reference fuse on layer 9** — done, exactly as you described.
2. **`backout_lambda` absorbed into MUDD** — the `self.scalars` slot is gone and the `x -= backout_lambda * x_7` line is gone. `dense_bs[1, 1, 1] = -4.34` puts an effective `≈ -0.5 * x_7` add at init into layer 10's residual through MUDD; the model re-tunes from there. Cleaner formulation, neutral-to-slightly-better loss in 10 seeds.
3. **MUDD-coef VE gate (your suggestion 4) — applied only to layer 10.** Layer 9's MUDD already produces a `(B, T, C, L)` tensor; v2 extends it to `(B, T, C, L+1)` and uses the extra `(L+1)`-th column as the per-token VE gate consumed by layer 10 (replacing `2 * sigmoid(F.linear([x[:6], ve[:6]], ve_gate_w))` for that one layer). Tiled across heads via `repeat_interleave`. **Why only layer 10:** I tried it on other layers too and it does help loss, but those layers don't have a MUDD `dense_w1` / `dense_w2` stack already, so each extra layer would need its own MLP+einsum and the per-step overhead ate the loss-per-step gain. Layer 10 is the special case where the MLP is already in the forward, so the marginal cost is essentially one extra einsum column.
4. **Bonus: pre-`lm_head` MUDD also mixes `ve[1]`** (the layer-1 VE bank). Same `dense_w1` → `dense_w2` path with one more `L` slot; nearly free.
5. **Code refactor.** Generic `mudd_layer_set` / `mudd_hidden_indices` / `hidden_append_set` loop is gone, replaced with two explicit `if i == N-2` / `elif i == N-1` blocks plus snapshot vars (`h7_snap`, `h9_snap`, `next_ve_gate`) and a docstring on `_init_mudd` listing sources per layer. Same arithmetic on the kept subset, fewer lines, top-to-bottom data flow.

Not in v2 (left for follow-ups):

- Your suggestion 3 (residual → V subspace projection) — would prefer to land it as a separate PR so the loss/time delta is attributable.
- Your suggestion 5 (MUDD-coef on the 3→6 skip).
- Mixing more / different VE banks into the pre-`lm_head` MUDD (only `ve[1]` for now).
- LR / betas for `dense_w1` / `dense_w2` / `dense_bs` are still untuned.

Happy to split anything out into a separate PR if you'd prefer.

varunneal:
Really cool! how did you find which layers were most important to write to the muddformer value residuals? Looks like there's just one read in the Value Muddformer graph (layer 10). Then we process `x` (also in Layer 10). Doing this all in one layer should allow for a big cleanup of `dense_bs` and `dense_w2`, which are mostly zeros. We just need to save the layer 9 residual for later to calculate  v_mudd, residual, ve_gate. I think we can bake in the value mudd res into `attn_args`.

Lisennlp:
> how did you find which layers were most important to write to the muddformer value residuals?

Empirically, by sweeping subsets of `{x0, h3, h5, h7, h9}` — the natural candidates given the existing architecture (`x0` = input skip, `h3` = source of the 3→6 skip, `h7` = the backout source, `h9` = the most recent post-MLP). In 5–10 seed sweeps, adding `h3` / `h5` cost wall-clock without reliable loss gains, and `{x0, h7, h9}` for layer 9 / `{x0, h9}` for layer 10 was the cheapest subset that still recovered the full quality gain. v2 then adds `ve_bank0 = ve[1]` to layer 10 because it was nearly free given the shared `dense_w1` stack.

I didn't run a serious "which residual stream is most informative" analysis — it was mostly "which of the architecture's existing residual snapshots are cheap to grab and route through MUDD". The fact that `h7` wins is partly because backout was already telling us h7 carries a useful "what to subtract before lm_head" signal.

> Doing this all in one layer should allow for a big cleanup of dense_bs and dense_w2, which are mostly zeros. We just need to save the layer 9 residual for later to calculate v_mudd, residual, ve_gate. I think we can bake in the value mudd res into attn_args.

Strong agree on both:

1. **Bake `v_mudd` into `attn_args`** is a clear win. Right now layer 9 → layer 10 uses a side-channel (`x` becomes the tuple `(x, v_mudd)`, and the per-layer loop has an `if v_mudd is not None:` branch that's dead code on layers 0–8). `precomputed_ve_gate` is already on `AttnArgs` for layer 9 → layer 10's gate; moving `v_mudd` next to it lets me drop the tuple-unpacking branch in `CausalSelfAttention.forward` and the side-channel branch in the per-layer loop.

2. **Consolidating to one `dense_w1` call on `h9_snap`** would clean up the all-zero slots: today `dense_w2` is `(num_mudd_layers=2, inter_dim=64, C=2, L+1=4)` = 1024 entries, of which `dense_w2[1, :, 0, :]` (layer-10 V channel, 256 entries) and `dense_w2[0, :, 1, _mudd_L]` (layer-9 ve_gate column on residual, 64 entries) are unused. The cleanup also drops the `num_mudd_layers` dim entirely — we'd have one `dense_w1` and one `dense_w2`, which is more honest: it's really one MLP gated against one snapshot.

The one subtlety I'd want to validate before merging: today layer 9's `dense_w1` reads `x = post-layer-9-MLP` and layer 10's `dense_w1` reads `x = post-layer-10-MLP` — different inputs. Consolidating means using `h9_snap` for both, so layer 10's residual delta loses the post-layer-10-MLP signal. I'd expect this to be neutral in 5–10 seeds (h9 and post-layer-10 aren't dramatically different in this regime), but worth a sweep so we don't regress loss when shrinking params.

ClassicLarry:
I have some syntax updates I'll make when merging, but waiting until 8H100 available on prime intellect. May be a couple days until H100s available.

Lisennlp:
Posted v3 (new commit + records under `records/track_1_short/2026-04-22_MuddFormer/this_pr_v3/`). README updated; headline:

|                | Steps | Time μ  | Loss μ | Loss σ | Loss p vs 3.28 |
| -------------- | ----- | ------- | ------ | ------ | -------------- |
| baseline-1480  | 1480  | 84.7039 | 3.2788 | 0.0011 | 0.0036         |
| this_pr_v2     | 1415  | 82.1289 | 3.2781 | 0.0010 | 0.0001         |
| **this_pr_v3** | **1415** | **81.9146** | **3.2780** | **0.0009** | **< 0.0001** |

v3 is **-2.79s vs baseline-1480** at strictly-better loss, and **-0.21s vs v2** at same step count with tighter variance (σ 0.0009 vs 0.0010). 10 seeds each.

What's in v3:

1. **6 dynamic per-token gates for layer 10, produced by layer 9's MUDD.** `dense_w2[0]` extended from `(inter_dim, C=2, L+1=4)` to `(inter_dim, C=2, L+5=8)`. The extra columns produce `resid_attn`, `post_attn`, `resid_mlp`, `post_mlp`, `x0_lambda`, and `bigram_lambda` — all for layer 10 — replacing the corresponding static per-layer scalars. Bias init matches the old static values so the model starts at the same effective coefficients and learns per-token modulation from there. All ride on the existing `dense_w1` → `dense_w2` einsum (extra columns only), so per-step overhead is negligible.

2. **`skip_connection` (layer-3 output) as a 5th source for post-loop MUDD.** Layer 10's residual source set is now `{x0, h7, h9, ve_bank0, skip_connection}`. The layer-3 hidden is a long-window snapshot that gives the final MUDD access to an early, broad-context representation. Same shared MLP path, one more `L` slot.

3. **Layer-10 MUDD moved post-loop.** The `elif i == N-1` branch is gone; computation happens once after the main loop exits. Pure refactor, same arithmetic, simpler compiled graph.

The philosophy is the same as v2: keep reusing the layer-9 MUDD MLP that's already in the forward to produce as many useful signals as possible for layer 10, so the marginal cost per new gate/source stays near zero. Making the layer-10 scalars input-dependent is a natural next step — these are the coefficients closest to the lm_head, so per-token modulation has the most leverage there.

Not in v3 (same list as before):

- MUDD on more than 2 layers.
- MUDD-coef on the 3→6 skip connection (your suggestion 5).
- VE-gate replacement on layers other than 10.
- LR / betas tuning for `dense_w1` / `dense_w2` / `dense_bs`.

ClassicLarry:
sorry about the delay here, Prime Intellect hasn't had 8h100's available for the last couple weeks. If still none available this weekend, I'll use a different provider to validate and merge.

ClassicLarry:
merging at 84.4-2.6 = 81.8. Will merge a slight refactor right after.

PR #264, opened 2026-04-29
New Record: XSA gated layers (-30 steps, -0.55s on 8xH200, -0.7s on 8xH100)

djdumpling:
# Gated XSA: per-(layer, head) learnable Exclusive Self-Attention

Add a learnable per-(layer, head) gate that subtracts the v-aligned component of each attention output, applied to every non-paired attention layer. It’s zero-initialised so the network is a strict superset of master at step 0, and XSA only kicks in if the optimiser decides it helps. The gain in loss headroom is spent on a 30-step shorter schedule (1440 to 1410), netting a ~0.7 s wall-clock improvement at iso-loss.

```bash
  baseline-1440    10   1440  80.400    0.105    0.0000      nan    3.27941  0.00145  0.00000    0.1151
  baseline-1410    10   1410  78.760    0.046   -1.6400    0.0000   3.28316  0.00093 +0.00375    1.0000
        this PR    10   1410  79.846    0.206   -0.5541    0.0000   3.27865  0.00105 -0.00076    0.0014
```

**Hardware: 8×H200**, not H100. All numbers above are H200, so they'll need re-validation on H100. The loss improvement is purely architectural and is hardware-independent. The wall-clock improvement decomposes into (a) the −30 scheduled steps, which save ≥ proportional time on H100 and (b) ~0.6 ms/step of added bandwidth-bound elementwise ops in attention (v normalize, dot-product with y, fused subtract — no new matmul or comms). The added ops scale with HBM bandwidth, so even with H100's ~30% lower bandwidth on bf16 activations, the per-step overhead stays well under a millisecond.

**XSA**: in a transformer attention block, each token's attention output `y` includes some component aligned with its own value vector `v`, and that self-aligned component carries information the token already knows. **Exclusive Self-Attention** ([Zhai,  2025](https://arxiv.org/abs/2603.09078)) removes it explicitly via `y ← y − ⟨y, v̂⟩ v̂` with `v̂ = v / ‖v‖` so the residual stream only receives information from *other* tokens through attention. The strength is per-head learnable via `y ← y − tanh(α_{layer,head}) · ⟨y, v̂⟩ v̂` with `α` a `(num_layers, num_heads)` parameter, **zero-initialized**. `tanh(0) = 0`, so at step 0 the model is bit-identical to master; the gate can grow to ~+1 (full XSA), shrink toward 0 (no-op), or even go slightly negative (mild self-amplification).

**Where XSA is applied**: only on **non-paired** attention layers. Paired layers ({0, 2, 5, 9}) reshape `v` across pairs of heads and the per-head projection math doesn't line up; layer 6 is MLP-only. That leaves XSA active on layers `{1, 3, 4, 7, 8, 10}` — 6 layers, all 12 heads each.

We also tried two narrower subsets and discarded both:
- `{7, 8, 10}` (the original `exp/xsa` set, but with the new gate): val 3.2782 ± 0.0009, no wall-clock benefit at the same step count.
- `{7, 10}` (drop layer 8, since its diagnostic α magnitudes were the smallest): no improvement over `{7, 8, 10}`.

All 12 heads in every active layer learned non-trivial positive α: layer 7 saturated near +0.9 across all heads, layer 8 hovered in +0.1 to +0.5, layer 10 in the middle. So the gate is genuinely being used.


djdumpling:
Both branches re-run on the same 8× H100 80GB HBM3 RunPod box on 2026-05-07. (still need Prime Intellect rerun), all runs are around 1.5 seconds off. 
                                                                                                                                                                                                  
```bash
 baseline-1440     7   1440  85.902    0.069    0.0000       nan   3.27987  0.00050  0.00000    0.2644                                                                                          
       this PR    11   1410  85.217    0.070   -0.6844    0.0000   3.27890  0.00118 -0.00097    0.0056
```                                                                                         
PR meets val_loss ≤ 3.28 with p<0.01 (one-sample t, n=11, t=-3.10) and is faster on wall-time, on one-sided PR<master, t=-20.36, df=13.1, p=1.36e-11, Δ=-0.684s.                                                                                                                          
                                                                                                                                                                                                  
Env vs Dockerfile: torch 2.10.0+cu128 (stable, not nightly cu126), kernels 0.13.0 (0.14 added a `trust_remote_code` check that rejects `varunneal/flash-attention-3` without `trust_remote_code=True`, which `train_gpt.py` doesn't pass), triton 3.6.0, Python 3.12.3, driver 580.126.09, CUDA 13.0.  Both branches sit ~+0.0005 higher in val_loss than the H200 sweep above, likely due to numerics drift in cu128/kernels-0.13, not anything specific to the change.

djdumpling:
XSA also ports nicely MUDD (I lost the logs unfortunately; this was run on same H200 cluster as before; not aformentioned RunPod nor Prime Intellect), but XSA+MUDD @ 1405 steps reduces val_loss by another ~0.0047 vs xsa gated alone, and is ~1.44 s faster on top of xsa-gated's −0.55 s win. One-sample t for the loss rule on these numbers: t as approximately −13.3, df=7, so effectively p=0.

ClassicLarry:
Looks good. 30 step decrease holds. Merging at 81.8-0.6 = 81.2s, retimed on latest branch.

ClassicLarry:
There are 6 heads per layer and not 12. Mentioning because if you were doing some analysis where you were seeing 12 learnable scalars, then something unexpected going on with the shapes.

Issue #263, opened 2026-04-29
Why not use DDP on Track_3_Optimization ?


alint77:
Since one of the goals of track3 is being more accessible, why not use DDP for better Comms overlap on PCIE setups? 

with manual NCCL comms, everything is on main gpu compute thread/stream, but using torch DDP allows some overlap by running NCCL calls on separate thread. 

here's a before and after: 

BEFORE: 

<img width="1472" height="864" alt="Image" src="https://github.com/user-attachments/assets/b51843da-4d32-47aa-adaa-d3ef096065cd" />

AFTER: 

<img width="1473" height="858" alt="Image" src="https://github.com/user-attachments/assets/3b386a92-b1ed-425f-a29e-36b2067c23e8" />

On this 4xA100 PCIE system saves 10% wallclock per step with same loss curve while simplifying the code even further at the same time @KellerJordan

alint77:
the code with DDP + profiling [is here](https://github.com/alint77/modded-nanogpt/tree/dev/ddp-profiler-dionNormuon) if anyone's interested

KellerJordan:
hm, on 4xA100, I'm seeing 628ms/step without DDP, and 624ms/step with DDP

alint77:
Is it NVlink or PCIe ?

PR #265, opened 2026-04-29
New Record: Track_3_optimization: Normuon for MLP + CWD 3300 steps (-200 Steps)

alint77:
## Summary

This PR adds a DDP Track_3_optimization training script (#263) and updates the optimizer setup for the optimization benchmark.

The main change is a hybrid optimizer:

- I'm using Muon and Normuon from the Microsoft Dion repo since it's much more optimized in terms of throughput and also I'm a contributor. 
- Attention Q/K/V/O matrices use Muon.
- MLP up/down matrices use NorMuon.
- Both Muon and NorMuon use cautious weight decay.
- MLP NorMuon uses a more aggressive learning rate than the original Muon baseline.
- The script uses Flash Attention 3 on H100 when the kernel is available, with PyTorch SDPA as a fallback.
- The script logs per-step timing and MFU estimates for H100 NVL.
- The script captures a short PyTorch profiler trace.

## Result

This run reaches the target before the final validation point:

- Step 3375: validation loss `3.27842`
- Step 3500: validation loss `3.27400`

The previous Track 3 Muon baseline reached:

- Step 3375: validation loss `3.28126`
- Step 3500: validation loss `3.27673`

figure

So this run beats the baseline at step 3375 and finishes with a lower final validation loss, I might be able to push it down to 3325 or so, might try later when have access to more compute.

## Optimizer Details

AdamW is still used for embeddings, the final projection, and scalar parameters.

Attention matrices use Muon with:

- `lr=0.025`
- `weight_decay=0.0125`
- `mu=0.95`
- `nesterov=True`
- `adjust_lr="spectral_norm"`

MLP matrices use NorMuon with:

- MLP up: `lr=0.030`, `weight_decay=0.025`
- MLP down: `lr=0.060`, `weight_decay=0.0125`
- `mu=0.95`
- `muon_beta2=0.9`
- `nesterov=True`
- `cautious_wd=True`
- `adjust_lr="spectral_norm"`

The motivation is to keep attention close to the strong Muon baseline, while applying NorMuon where it is most likely to help: the large rectangular MLP matrices.


alint77:
**UPDATE:**

Tuned WD, doubling it compared to baseline since its using Cautious WD. brought down the loss at 3500 to ~3.265. 

Ran 3 times for 3300 steps, all under 3.28. 

Also tried to get per-head-NS to work but no matter the hparams, it performed worse. 

figure

KellerJordan:
Hello, thanks for this work. To ensure statistical power I've had to add the following rule: we require (3.28 - mu) * n**0.5 >= 0.004, where mu is the average result over n non-cherry-picked runs

alint77:
Thanks! I will rerun as soon as I have access to more compute.

KellerJordan:
gave shared credit for new Muon hparams (double wd) https://github.com/KellerJordan/modded-nanogpt/commit/d035e686c3fab8e5f8f72515fe3f4c2932911022#diff-551227b77b683c9ea50df09ee5424d491fa740cbc6bb6aa3f2127142172d1b3fR30

PR #274, opened 2026-05-02
New Record: Track_3_optimization: Skylight 3250 steps (-250 Steps)

kumarkrishna:
## Summary

3250 steps to <3.28 val loss with Muon + NorMuon-lite + u/w-floor, beating the current MuonH SOTA at #5 by 250 steps. Supersedes #269 (closed before the new Rule 3/4 update; this submission complies with both).

## Mechanism

On top of standard Muon (Newton-Schulz orthogonalization + aux Adam):

- **Per-row variance EMA on the NS update** (NorMuon-lite — port of the track_1 2025-10-24 entry): for each row of the orthogonalized update, maintain an EMA of squared values; divide each row by sqrt(EMA); renormalize Frobenius norm back to the original. Spectral structure preserved; rows that were over- or under-weighted by NS get adaptive rescaling.
- **u/w-floor**: when `||u||_F / ||w||_F < 0.35`, scale up the update so the ratio equals 0.35. Acts in place of weight decay's role of keeping update magnitude proportional to weight magnitude (`wd` is then small / fine-tunable independently).
- **Hyperparams**: `lr=0.0375`, `wd=0.025`. AdamW-side params unchanged from baseline.

## Statistical evidence (Rule 3)

- n=8 runs (4 H100 nodes × 2 cycles, 8 GPUs each)
- per-run val_loss: 3.27543, 3.27615, 3.27651, 3.27726, 3.27770, 3.27775, 3.27787, 3.27794
- mu = 3.27708
- Margin: `(3.28 - 3.27708) * sqrt(8) = 0.00827` ≥ 0.004 ✓

## Compliance with the new rules

- **Rule 1**: same dataset / batch / architecture as baseline ✓
- **Rule 2**: single forward-backward per step; val_loss is logged but never read for any decision ✓
- **Rule 3**: see above ✓
- **Rule 4**: embedded train-script source in each logfile is fully self-contained — no third-party imports ✓ (diagnostic instrumentation present in #269 has been stripped; it had no effect on training numerics)

## Files

- `records/track_3_optimization/README.md`: +1 row
- `records/track_3_optimization/results/20260501_skylight001/`: 8 logfiles

No changes to the baseline `train_gpt_simple.py`, `figure.png`, or `make_figure.py`.


lliu606:
I checked the code and found that Skylight doesn't use weight decay?  Maybe it has some connection with NorMuonH?

KellerJordan:
I ended up retroactively lowering the stat sig threshold to `(3.28 - mu) * n**0.5 >= 0.004`, so even the closed earlier PR has become valid. Will accept this, with the earlier date listed in the history to reflect that

KellerJordan:
One point that I need to modify is that NorMuon-lite here seems to be the same as a regular NorMuon. Since the normal NorMuon ([arxiv](https://arxiv.org/abs/2510.05491)) is already doing Adam-style 2nd moment per-row rather than per-element. Please let me know if I have this wrong.

KellerJordan:
I am also removing the listed weight decay hparam from the result description, since like @lliu606 pointed out, it is unused

PR #290, opened 2026-05-08
Track 3: KL-SOAP-H: 3125 steps (n=6)

WhenWen:
## Summary
- Update the Track 3 KL-SOAP-H result to the 3125-step clean sweep.
- Replace the prior 3150-step logs with the new clean sweep logs.
- Regenerate the benchmark overview figures.

paper link https://arxiv.org/abs/2509.03378
If one precondition every step, seems like it is no longer necessary to do NS iteration on top of it

## Result
- Passing hyper: beta1=0.95, beta2=0.9, shampoo_beta=0.9, precondition_frequency=1, lr=0.018.
- n=6 mean validation loss at 3125 steps: 3.278003.
- Range: 3.27706-3.27895.
- Significance check: `(3.28 - mu) * sqrt(n) = 0.004891 >= 0.004`.
- The first requested tuple, beta1=0.9 and shampoo_beta=0.9, was also logged and did not pass: mean 3.28294.

## Cleanliness
- Logfiles contain a no-args `train_gpt_simple.py` version derived from current master.
- Changes in the logged training script are localized to optimizer/init/hyperparameter sections plus fixed single-trial execution.
- The only external seed control is `KL_SOAP_SEED`; optimizer hyperparameters are hardcoded in the logged script.

## Verification
- `python3 make_figures.py`
- `git diff --check`
- Parsed committed logs for the 3125-step mean/significance check.
- Validated embedded log code blocks parse and contain the expected hardcoded KL-SOAP-H tuple.


lliu606:
Cool! I'm wondering that why it's KL-SOAP-H rather than SOAP-H?

WhenWen:
> Cool! I'm wondering that why it's KL-SOAP-H rather than SOAP-H?

KL Soap paper found this update rule to be more stable and I really like the theoretical intuition. I haven't tried Soap in this track at all so don't really know the performance

lliu606:
I just submitted the result for a SOAP based variant #298. Based on the result, I guess pure SOAP and SOAP (factorized), as described in the [SOAP paper](https://arxiv.org/abs/2409.11321)￼, may be worth trying :)

PR #295, opened 2026-05-13
Track 3: Normalized Correction Reduce Muon 50 steps for free (3325, No hyperparameter tuning required, no extra memory needed, almost no computational cost)

Yuan-Jinghui:
We have taken note of the perspective presented in Aurora @liyang2019 : if the update amount $\Delta W$ is not constrained, it can lead to neuron inactivation and affect efficiency. They further point out that this is caused by the differing $\ell_2$ norms of the matrix rows fed into the Newton-Schulz (NS) iteration, which serves as our primary motivation.

Inspired by [Nora,](https://arxiv.org/abs/2605.03769) we attempted to correct the momentum $m$ before the NS iteration by left-multiplying it by $\text{diag}(mm^T)^{-1/2}$, thereby ensuring that every row of $m$ has the same $\ell_2$ norm. (We also highly recommend reading Nora, which is a very interesting piece of work.)

However, the operating environments of Nora and Muon differ. Directly using $\text{diag}(mm^T)^{-1/2}$ to correct $m$ eliminates the magnitude differences between the rows of $m$, which negatively impacts the NS iteration. (Nora remains applicable as it does not require NS iteration). To address this, we use $\text{diag}(mm^T)^{-1/4}$ for left-multiplication and, for energy balance, use $\text{diag}(m^Tm)^{-1/4}$ for right-multiplication.

Specifically, before passing $m$ into the NS iteration, we apply a technique we call "Normalized Correction":

$$m = \text{diag}(mm^T)^{-1/4}\  m\  \text{diag}(m^Tm)^{-1/4}$$

The implementation requires only a few lines of code:

```python
def muon_update(grad, momentum, mu=0.95, nesterov=True):
    momentum.lerp_(grad, 1 - mu)
    update = grad.lerp(momentum, mu) if nesterov else momentum
    r_norm = update.norm(dim=-1, keepdim=True)
    c_norm = update.norm(dim=-2, keepdim=True)
    update = update / torch.sqrt(torch.clamp(r_norm * c_norm, min=1e-12))
    update = zeropower_via_newtonschulz5(update)
    update *= max(1, grad.size(-2) / grad.size(-1))**0.5

```

This code requires no additional storage, involves almost no extra computation, and is plug-and-play.

To ensure fairness, we did not tune the hyperparameters and used the default settings from the 6th record @nilin @alint77. Even so, the Normalized Correction demonstrated a competitive improvement.

Due to limited GPU resources, we only conducted a single run. To maintain statistical significance, we have made a conservative claim of a 50-step improvement. In reality, the model first reached approximately 3.27837 at step 3275, and by the 50-step mark we cited, it had already reached 3.27552.

My friends and I all contributed to the Normalized Correction, including: Jinghui Yuan (myself), Jiaxuan Zou @JiaxuanZou0714, and other authors of Nora (Shuo Wang @SIRBabbage ...).

### DIFF
--- a/records/track_3_optimization/results/20260513_muonhn/688740c6-db37-41d4-bcd6-58528b342b23.txt
+++ b/records/track_3_optimization/results/20260513_muonhn/688740c6-db37-41d4-bcd6-58528b342b23.txt
@@ -1,0 +1,443 @@
+"""
+train_gpt_simple_muonh.py
+
+This file descends from the [NanoGPT speedrun](https://github.com/KellerJordan/modded-nanogpt).
+It was prepared as a simplified version of the speedrun for use in neural net optimization research.
+
+Differs from `train_gpt_simple.py` only in the init and the optimizer: the matrix-parameter
+Muon update is replaced by MuonH (the same Newton-Schulz orthogonalised direction, applied
+via a hyperball projection that preserves the Frobenius norm of every hidden 2D weight
+matrix at every step), and the residual-side projections / mlp.fc are initialised with
+per-module multipliers on the default Kaiming init so MuonH has non-zero matrices to operate
+on from step 0.
+"""
+
+import os
+import sys
+with open(sys.argv[0]) as f:
+    code = f.read() # read the code of this file ASAP, for logging
+import uuid
+import time
+from pathlib import Path
+
+import torch
+from torch import Tensor, nn
+from torch.optim import AdamW
+import torch.nn.functional as F
+import torch.distributed as dist
+
+
+########################################
+#              Dataloader              #
+########################################
+
+def _load_data_shard(file: Path):
+    header = torch.from_file(str(file), False, 256, dtype=torch.int32) # header is 256 int32
+    assert header[0] == 20240520, "magic number mismatch in the data .bin file"
+    assert header[1] == 1, "unsupported version"
+    num_tokens = int(header[2]) # number of tokens (claimed)
+    with file.open("rb", buffering=0) as f:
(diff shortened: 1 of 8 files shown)


Yuan-Jinghui:
### **Normalized Correction consistently brings improvements!**

This time, we tested the configuration of record [12,](https://github.com/KellerJordan/modded-nanogpt/blob/master/records/track_3_optimization/results/1bd8db7a-f3a3-4195-856d-cab7e0816443.txt) still without tuning hyperparameters, increasing memory footprint, or adding significant computational overhead. Due to GPU constraints, we are reporting conservative results: Normalized Correction reduces 12 by 25 steps. We conducted two runs, and since $(3.28 - (3.27605 + 3.27711) / 2) \cdot \sqrt{2} > 0.004$, it appears that Normalized Correction offers even greater gains; however, we report the conservative figures here.

We will subsequently test the performance of Normalized Correction on MuonH, which may be of interest to @whenwen.

Future research will continue, and we will release a blog post or an arXiv preprint.

`
Contributors={Jinghui Yuan and Jiaxuan Zou and Shuo Wang and Yong Liu and Feiping Nie}
`

Yuan-Jinghui:
### **Normalized Correction Improves MuonH for Free (-25 steps)**

We selected parameters perfectly identical to MuonH, and the number of steps required to reach 3.28 was reduced by 25, currently standing at 3300 steps. The confidence calculation is as follows:
 $(3.28 - (3.27704 + 3.27833 + 3.27675 + 3.27835 + 3.27900) / 5) \cdot \sqrt{5} > 0.004$. 

We look forward to Normalized Correction bringing even greater progress.

@Whenwen

PR #298, opened 2026-05-14
Track3: Add SinkSOAP optimizer result: 3090 steps

lliu606:
### Summary

This PR adds a new optimizer variant, SinkSOAP, which reaches the target at 3090 steps.

SinkSOAP is a [SOAP](https://arxiv.org/pdf/2409.11321) based variant inspired by [DeepSeek MHC](https://arxiv.org/pdf/2512.24880) and [SinkGD](https://arxiv.org/pdf/2502.06742). Instead of using the second-order moment **V** as the momentum preconditioner in the SOAP rotated space, SinkSOAP applies a Sinkhorn-style row/column normalization directly to the rotated momentum.

### Method

Let **M** denote the momentum update (with nesterov), and let $Q_L, Q_R$ be the left and right rotation matrices as denoted in [SOAP](https://arxiv.org/pdf/2409.11321). We first rotate the momentum into:

$\tilde{M} = Q_L M Q_R$

Unlike standard SOAP-style second-order preconditioning, SinkSOAP does not maintain or apply a second-order moment in this rotated space. Instead we then apply Sinkhorn-Knopp iteration to $\tilde{M}$ so that the update norm is more evenly distributed across rows and columns.

A simplified iteration is:

```math
\tilde{M}^{(t + 1/2)}_{ij}
=
\frac{\tilde{M}^{(t)}_{ij}}
{\sqrt{\sum_k |\tilde{M}^{(t)}_{ik}|^2 + \epsilon}}
```

```math
\tilde{M}^{(t + 1)}_{ij}
=
\frac{\tilde{M}^{(t + 1/2)}_{ij}}
{\sqrt{\sum_k |\tilde{M}^{(t + 1/2)}_{kj}|^2 + \epsilon}}
```

Finally, we rotate the normalized update back to get the final update:

$O = Q_L^\top \tilde{M}^{(T)} Q_R^\top$

In the submitted run, we use $T = 10$ Sinkhorn-Knopp iterations. Empirically, however, a single iteration already works well, and increasing to 10 iterations only provides a small additional gain.

This makes the method Adafactor-like in spirit, especially in the single-iteration case. However, SinkSOAP is not the same as SOAP (factorized) from [SOAP](https://arxiv.org/pdf/2409.11321)￼, since it does not use any second-order term as the main preconditioner.

### Handling non-square 2D hidden matrices

SinkSOAP is mostly an independent optimizer and is not mixed with most previous methods from other PRs, including Muon. The only inherited component is the size-dependent normalization idea from NorMuon.

For 2D hidden matrix parameters where `size(0) != size(1)`, we follow [NorMuon](https://arxiv.org/pdf/2510.05491) by maintaining an EMA second-order moment only along the shorter dimension. This moment is then used to normalize the SinkSOAP update $O$, producing a more row/column-balanced update.

Our ablation shows that this NorMuon-style normalization just provides a small (but consistent) improvement over pure SinkSOAP.

### Hyperparameters

- train_steps: 3125
- lr: 0.04
- first order momentum $\beta_1$: 0.95
- SOAP $\beta_2$: 0.95
- NorMuon $\beta_2$: 0.95
- weight_decay: 0.025
- lr_schedule: WSD with `cooldown_frac=0.7`, same as Muon.
- Sinkhorn-Knopp iterations: 10

logfile: `records/track_3_optimization/results/20260514_sinksoap/d0155dd0-f77d-48a9-8eb4-453f894b9476.txt`

### Result

The result is evaluated over 10 runs. At step 3090, the result is:
- mean loss: 3.278464
- Significance check:  `(3.28 - 3.278464) * sqrt(10) =0.004857 > 0.004 `

### Notes

SinkSOAP suggests that, using a second-order-moment preconditioner in the rotated SOAP space may be unnecessary or even suboptimal. A simple Sinkhorn-style normalization of the rotated momentum already gives strong performance.

Apart from the NorMuon-style normalization for non-square 2D hidden matrices, SinkSOAP remains a clean standalone optimizer.

---

Thanks to @zhenghaoxu-gatech for running the experiments.

### DIFF
--- a/records/track_3_optimization/results/20260514_sinksoap/d0155dd0-f77d-48a9-8eb4-453f894b9476.txt
+++ b/records/track_3_optimization/results/20260514_sinksoap/d0155dd0-f77d-48a9-8eb4-453f894b9476.txt
@@ -1,0 +1,1088 @@
+"""
+train_gpt_simple.py
+
+This file descends from the [NanoGPT speedrun](https://github.com/KellerJordan/modded-nanogpt).
+It was prepared as a simplified version of the speedrun for use in neural net optimization research.
+"""
+
+import os
+import sys
+with open(sys.argv[0]) as f:
+    code = f.read() # read the code of this file ASAP, for logging
+import uuid
+import time
+from pathlib import Path
+
+import torch
+from torch import Tensor, nn
+from torch.optim import AdamW
+import torch.nn.functional as F
+import torch.distributed as dist
+
+
+########################################
+#              Dataloader              #
+########################################
+
+def _load_data_shard(file: Path):
+    header = torch.from_file(str(file), False, 256, dtype=torch.int32) # header is 256 int32
+    assert header[0] == 20240520, "magic number mismatch in the data .bin file"
+    assert header[1] == 1, "unsupported version"
+    num_tokens = int(header[2]) # number of tokens (claimed)
+    with file.open("rb", buffering=0) as f:
+        tokens = torch.empty(num_tokens, dtype=torch.uint16, pin_memory=True)
+        f.seek(256 * 4)
+        nbytes = f.readinto(tokens.numpy()) # avoid bytes->array copy
+        assert nbytes == 2 * num_tokens, "number of tokens read does not match header"
+    return tokens
+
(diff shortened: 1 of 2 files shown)


WhenWen:
Personally found it very surprising that this work better than Muon

PR #299, opened 2026-05-14
Track 1 (new record, -1.0s): sign-trick to better compress bigram embeddings

trianxy:
This PR improves the track 1 (small) by ~1.0s by implementing the below bullets 1 & 2 around bigrams.

**Results**
|  | baseline (n=10), main branch on May 20, 2026 (3546294c065bef9fc351e1d34d11c542b5c98f70)  | commit https://github.com/KellerJordan/modded-nanogpt/pull/299/changes/647c762172d57c828a9d9be6e3105b2cd775dd03 of this pr299 (n=20) | delta |
|---|---|---|---|
| val_loss mean | 3.27937 | 3.27917 | -0.00020 |
| train_time mean (ms) | 85394.3 | 84429.5 | **-964.8** |

p-value that val_loss of this PR is <3.28 is 0.0024 (calculated using code [here](https://github.com/KellerJordan/modded-nanogpt/tree/master/records/track_1_short/2025-01-04_SoftCap))

**Details**
1. [gain: ~0.5s] use sign-trick to better compress multiple bigrams' embeddings into a shared bigram_embed row. Let me explain:
    - Each embedding is shared across dozens of bigrams as ~50k*50k bigrams are mapped into fewer rows
    - Use the following 'trick' (Chen et al. 2015, "Compressing Neural Networks with the Hashing Trick") to allow 1 embedding to still carry a distinguishable meaning for each of its bigrams:
        - Suppose `g_j` is (bigram `j`)'s optimal embedding, i.e. the learned vector if there were no hash collisions with other bigrams
        - In the forward pass, multiply the embedding with a random `+/-1` sign pattern `s_j` (fixed for each `j`; is a vector of length `args.bigram_dim`). The learned embedding will become `g_j*s_j`. In the forward pass we use `(g_j*s_j)*s_j=g_j` and thus end up using the optimal embedding `g_j` -> multiplying with `s_j` didn't change what the forward pass effectively computes!
        - Now, if there are other bigrams i sharing the same embedding, we end up with a learned embedding `Σ_i(g_i*s_i)`. However, in the forward pass, we 'recover' j's optimal embedding because the cross-terms `(s_i*s_j)` for `i≠j` roughly cancel each other out (since we picked random signs), leading to `(Σ_i(g_i*s_i))*s_j ~ g_j` 
        - In other words, we 'hide' 50k*50k optimal embeddings g_j inside fewer rows (note: we could draw `s_j` from a Sylvester-Hadamard matrix for better cancellation but results were similar)

2. [gain: ~0.5s] instead of `(5*vocab_size, 768 dim)` use `(15*vocab_size, 192 dim)`
    - I found that reducing the bigram dimension by a factor of 2-8 (while increasing the bigram_vocab_size proportionally) helped. It seems the extra dims aren't used that much. And by reducing the bigram dimension, we reduce the number of floats which need to be moved when doing `self.bigram_embed(bigram_input_seq)` (-> bandwidth gains)
    - I did try using a linear layer to upscale (or just repeat) the 192 dim to end up with 768 dim but that didn't give any clear gains

## Negative results

3. Use trigrams
    - Adding these reliably improves the validation loss, but the cost to the runtime was too high, even when trying different tricks (e.g. have one embedding table for both bigrams and trigrams, and reuse some of the computations)
    - POSSIBLE FUTURE IMPROVEMENT: I think there should be a smart way to implement (some beneficial subset of) trigrams with very little additional runtime, but I wasn't able to achieve this during the afternoon I gave myself for this
  

4. bigrams: only consider top 100k bigrams (or top 1M), out of the 28M unique ones
    - Result: runtime didn't improve enough to offset the amount that validation loss got worse
    - Background: the most common bigram is [".","\n"] which appears ~1% of the time! In total, there are ~28m unique bigrams (in the first 400m training tokens). Note that theoretically, we could have had ~50k**2 = 2500M bigrams, but we have 100x less than that (thankfully)
    - it seems we benefit from taking all the bigrams: we aren't gaining too much from skipping infrequent bigrams (it seems that noise introduced by having multiple bigrams share the same slot is strongly related to their occurance, so the cost of having one more bigram share an already occupied slot is not big, if that additional bigram isn't common)

5. bigrams: VIP buckets for most common bigrams (e.g. top 16k bigrams get their own personal slot, i.e. they don't share a slot with other bigrams)
    - Didn't help validation loss; again, it seems that we don't loose too much by having multiple bigrams share the same slot
  
6. bigrams: reduce number of bigram slots
    - Based on 1 & 2, it should have been beneficial (meaning: no decrease in validation loss while reducing wall-time) if we strongly reduced the number of rows in model.bigram_embed. This turned out to be false. **This means, I do NOT understand the dynamics well of bigrams who share the same slot, since this behavior is not really in line with e.g. bullets 1 & 2 of this PR**
    - POSSIBLE FUTURE IMPROVEMENT: I believe there should be a way to get much more here (and THEN also be able to add trigrams and other ngrams)
  
7. PyTorch 2.11 and PyTorch 2.12: I tried both but they each gave no improvement (they seem to give similar val_loss but are ~0.5s slower)

## Other

- For my and other's future reference: currently, we can make the runtime ~0.3s faster by sacrificing ~0.001 val loss
~- Pinned kernel dependency to 0.13.0 because the new 0.14.0 requires changing the way we load FlashAttention (I saw @varunneal is looking into this anyway)~

_Edit1: changed format of explanation in bullet 1_
_Edit2: added results of the baseline runs vs the new runs_
_Edit3: I merged the current master branch (which included MUDD) in the latest commit [3351c0e](https://github.com/KellerJordan/modded-nanogpt/pull/299/commits/3351c0eb2a284f03ce344027068d4b54037eb7df). However, I didn't update the loss/runtime numbers in the above table because of a lack of 8h100s_


trianxy:
Keeping this PR as draft right now, because I haven't been able to get a hold of an 8h100 to create some clean logs (results are pretty much confirmed by a bunch of misc experiments I ran).

Once I do, I'll add the log files here and change from Draft to Ready for review.

varunneal:
I rlly like this PR. Heads up you can try switching to `get_kernel(kernels-community/flash-attn3)` hopefully that will work w latest kernels version

also instead of the long block comment in the code you can also link to this PR or paper 

also wondering if you tried any other involutions. E.g. instead of `g -> -g`   you can try `g -> 1 - g` or even bit manipulation (xor) ideas.

trianxy:
>also instead of the long block comment in the code you can also link to this PR or paper

Good point, will do.


> also wondering if you tried any other involutions. E.g. instead of g -> -g you can try g -> 1 - g or even bit manipulation (xor) ideas.

No, I didn't. But it's an Interesting idea! Other involutions might better reduce interference.

Initially, I thought just using the signs is pretty optimal. But as I point out in my bullet 6 in the [PR description](https://github.com/KellerJordan/modded-nanogpt/pull/299#issue-4447653173) , there's more to understand about how to best "compress" (and then "decompress") information of different bigrams into the same embedding vector.

Maybe I also need to think a bit harder about how the noisy interference impacts the backward pass (incl. interplay with how Adam updates the weights).

Let me know if you want to think through this together.

varunneal:
yeah I'm interested in getting this to work on our value embeddings too. Actually, I'm wondering if we can push it so far that we can make it work on the LM Head. Maybe with a learnable involution?

trianxy:
> yeah I'm interested in getting this to work on our value embeddings too. Actually, I'm wondering if we can push it so far that we can make it work on the LM Head. Maybe with a learnable involution? 

Learnable would be great. Will try to look into this in 1 week.

I d love to better understand mathematically (and feel) how information is stored inside those embedding vectors. And how learning happens.

This would allow using fewer, or lower-dimensional vectors, to store the same information. Unless we are already in an optimal regime. Which, at least for bigrams, we were not according to this PR

trianxy:
This PR is now ready for review. I ran the baseline 10 times and this PR 20 times. The good news is that the observed speed gain slightly improved from 0.7s to 1.0s.

trianxy:
> I rlly like this PR. Heads up you can try switching to `get_kernel(kernels-community/flash-attn3)` hopefully that will work w latest kernels version
> 
> also instead of the long block comment in the code you can also link to this PR or paper

FYI I used `get_kernel(kernels-community/flash-attn3)` and it worked well. Also, I removed the long block comment and am now linking to this PR.

xor2k:
> I rlly like this PR. Heads up you can try switching to `get_kernel(kernels-community/flash-attn3)` hopefully that will work w latest kernels version
> 
> also instead of the long block comment in the code you can also link to this PR or paper
> 
> also wondering if you tried any other involutions. E.g. instead of `g -> -g` you can try `g -> 1 - g` or even bit manipulation (xor) ideas.

Hi everybody, I heard xor and so I showed up :sweat_smile: Not really, but have a look at my PR: https://github.com/KellerJordan/modded-nanogpt/pull/314 There, I suggest PyTorch's built-in attention as a fallback. But maybe it can even be default? Can't benchmark it on H100s though, just got two DGX Spark here.

If you want to replace that kernel with kernels-community/flash-attn3: it won't work on aarch64, just on x86, see https://huggingface.co/kernels-community/flash-attn3/tree/main/build

ClassicLarry:
I tested on the latest version, that includes learnable XSA and mudd.

Before:
step:1385/1385 val_loss:3.2784 train_time:81700ms step_avg:58.99ms
step:1385/1385 val_loss:3.2771 train_time:81717ms step_avg:59.00ms

After, excluding new FA3:
step:1385/1385 val_loss:3.2798 train_time:79962ms step_avg:57.73ms
step:1385/1385 val_loss:3.2842 train_time:79962ms step_avg:57.73ms
step:1385/1385 val_loss:3.2792 train_time:80144ms step_avg:57.87ms
step:1385/1385 val_loss:3.2807 train_time:80047ms step_avg:57.80ms
step:1385/1385 val_loss:3.2801 train_time:80074ms step_avg:57.81ms

including new FA3 (appears no change)
step:1385/1385 val_loss:3.2797 train_time:79806ms step_avg:57.62ms
step:1385/1385 val_loss:3.2796 train_time:79904ms step_avg:57.69ms
step:1385/1385 val_loss:3.2790 train_time:79976ms step_avg:57.74ms
step:1385/1385 val_loss:3.2816 train_time:79811ms step_avg:57.63ms

Adding 10 steps:
losses, times:
(array([3.2759, 3.278 , 3.2771, 3.2759, 3.2781]),
 array([80.377, 80.478, 80.461, 80.382, 80.422]))

Will merge with +10 steps and -1.3s

ClassicLarry:
Also for future PRs, dont need to include baseline runs in the PR. and ideally if PRs are built on top of the prior open PR, it helps merges go faster through peer validation.

ClassicLarry:
In my test I accidentally moved the bigram embed injection point to after mlp instead of after attn. will recheck. this will end up a record in either case.

ClassicLarry:
Looks good as is. -1.5s.

ClassicLarry:
Applying the sign trick here is genius. Also really like the thoughts in the negative experiments.

I think there's a couple ways to try and interpret what the sign trick is doing. I am skeptical that we 'recover the optimal embedding'. We still only have 768 degrees of freedom for a bigram embed, which has to fit however many bigrams hash to the bucket. 

If I consider the simplest case with 2 bigrams that hash to the same bucket, we can now no longer give them the same effective embedding. On roughly half of their axis, they will be forced to have reflected representations of each other.

Maybe an even simpler case, is if we had 1 degree of freedom to represent two points on a line. We could specify a single scalar that both points sit at. Or, we could define some function f(p), that maps a point to the line, where that function has 1 learnable parameter. And in this case we have chosen to make that function m*sign(hash(p)), where m is the learnable parameter. However we could use any arbitrary function here instead of sign(hash(p)). I see you mentioned you tried 'Sylvester-Hadamard matrix'. It is hard for me to reason about how much the 'cancellation effect' matters. Anyways cool stuff!

PR #302, opened 2026-05-18
Track 3: SOAP-H: 3125 steps (n=6)

Fullfix:
## Motivation

I wanted to create a good SOAP baseline, so future SOAP-based modifications can be compared against plain SOAP more directly.

## Result

The run reaches mean loss `3.278175` over `n=6` seeds, with range `3.27696-3.27895`.

It passes the significance test:

```text
(3.28 - 3.278175) * sqrt(6) = 0.00447032 >= 0.004
```

## Key Parameters

Setup from the previous KL-SOAP-H result, with KL-SOAP replaced by ordinary SOAP and the LR schedule tuned.

Parameters: `train_steps=3125`, `lr=0.018`, `beta1=0.95`, `beta2=0.9`, `precondition_frequency=1`, no bias correction, `lr_steps=3200`, `lr_power=1.25`, `lr_floor=0.0`.

Seed is controlled by the `SOAP_SEED` environment variable and logged by the script.


varunneal:
I'm quite surprised that SOAP is so competitive, with this baseline beating most of the current records! Do you know how important/useful hyperball was here?

Fullfix:
Thanks for the question! I initially tried running SOAP without hyperball and with untuned scheduler, but it took way more than 3400 steps. And transfering the hyperball setup from @kaiyue-wen runs gave significant improvements with same hyperparameters, leading to ~3200 zone. Therefore I decided to stick with hyperball.

PR #306, opened 2026-05-20
Record: Track 1 (-0.8s) - FP8 MLP Up Projection Forward

msisovic:
## Summary

Pretty simple idea, quantize weights once per optimizer step, and activations once per forward pass, using per-tensor scales to FP8, thus utilizing the 2x FP8 FLOPs of H100 compared to BF16. A similar thing was done with the LM head already, but that one benefits more, as it's a larger matrix and thus is more compute-bound. 

Due to the slight loss of accuracy, if the same amount of steps is kept, we actually regress the final loss by `0.00082`, so I added 5 extra steps to compensate for this, achieving the same loss as the baseline, and observing the `0.8s` speedup from the title. Details in the "Validation" section.

A few things I tried but ended up not working, maybe someone can execute on them better:
- FP8 for the MLP down-projection, but its 4x wider input activations made the required scaling/casting pass too memory-bandwidth-intensive to pay off, even with delayed scaling to remove the amax-reduction cost.
- per-row scaling for better up-projection precision, but the _scaled_mm per-row-scale path regressed performance versus the simpler scaling variants.

## Validation

Notably, my runs didn't clear the `p<0.01` bar, but neither did the baseline reruns I did on the same machine. Thus I believe this PR is valid, but if the maintainers want to wait for the baseline to clear the threshold more comfortably before merging this, I'm fine with that. The loss is pretty much the same as baseline on the same machine, verified by the following script:

```python
import scipy.stats
import torch

baseline_losses = [3.28, 3.2787, 3.2783, 3.276, 3.2787, 3.2795, 3.2792, 3.2799]
baseline_times = [84.604, 84.677, 84.776, 84.513, 84.576, 84.684, 84.631, 84.618]

this_record_losses = [3.2802, 3.28, 3.2789, 3.2763, 3.2792, 3.2793, 3.2781, 3.2785]
this_record_times = [83.855, 83.75, 83.871, 83.9, 83.852, 83.901, 83.937, 83.763]

for name, losses, times in [
    ("baseline", baseline_losses, baseline_times),
    ("this_record", this_record_losses, this_record_times),
]:
    print(name)
```

Script output:
```
baseline
p=0.0156
losses: (tensor(0.0013), tensor(3.2788))
time: (tensor(0.0790), tensor(84.6349))

this_record
p=0.0148
losses: (tensor(0.0012), tensor(3.2788))
time: (tensor(0.0661), tensor(83.8536))
```


trianxy:
> we actually regress the final loss by 0.0038

Nice! By the way, I guess you missed a zero here and meant `0.00038` ?

msisovic:
> > we actually regress the final loss by 0.0038
> 
> Nice! By the way, I guess you missed a zero here and meant `0.00038` ?

Ha, thanks for noticing, it's actually neither! I was looking at the wrong logs (not included in the PR). It was `0.00082`. I edited the description accordingly.

trianxy:
> Ha, thanks for noticing, it's actually neither! I was looking at the wrong logs (not included in the PR). It was 0.00082. I edited the description accordingly.

Nice, that fits even better with my heuristic of 0.3s~0.001 val_loss, because 5 steps is roughly 0.28s which translates to ~0.0009 val loss

ClassicLarry:
Cool, nice change. Loss should be within the window now, I'll confirm.

ClassicLarry:
Merging at 79.2 (-0.5s), based on applying to latest branch with +5 steps.

yiakwy-xpu-ml-framework-team:
@ClassicLarry @msisovic  Note the performance for H800x8 is not good. I am investigating cause later

截屏2026-06-30 20 51 04

I will use this as baseline.

Thank you!

ClassicLarry:
> @ClassicLarry @msisovic Note the performance for H800x8 is not good. I am investigating cause later
> 
> 截屏2026-06-30 20 51 04
> I will use this as baseline.
> 
> Thank you!

I am guessing that you got a bad node, which sometimes happens. Both msisovic and myself got runtimes around 80s.

PR #300, opened 2026-05-14
Track 3: Aurora-on-mlp.proj + extended Contra-Muon on PR #294 stack (bin=2930, n=16)

eliebak:
## Summary

Adds a Track 3 optimization result: **bin = 2930 steps to 3.28 val_loss**, validated over **n=16 non-cherry-picked seeds (0..15)**.

This builds on [PR #294 (radial brake)](https://github.com/KellerJordan/modded-nanogpt/pull/294) and applies three changes on top of its recipe:

1. **Aurora row-balanced polar on `mlp.proj`** ([Tilde Research, May 2026](https://github.com/tilde-research/aurora-release)). In `zeropower_via_newtonschulz5`, when the matrix is wide (`m < n` — exclusively the 12 `mlp.proj` 768×3072 weights), run K=3 outer Aurora iterations with diagonal row-rescaling between Newton-Schulz calls (β=0.25, ε=1e-7). Square (attn QKV/proj) and tall (`mlp.fc`) matrices use plain MuonEq + Polar Express NS-5 unchanged. Aurora is applied only to the wide direction because on tall matrices the row-uniform constraint fights the depth-scaled `mlp.fc` init.
2. **Soft-Muon disabled** (`SOFT_MUON_CEIL: 0.80 → 0.00`) and **NorMuon-lite disabled** (`NOR_BETA2: 0.95 → 1.0`). The stack-contribution figure in the result directory measures Δ val_loss when each component is removed from the intermediate stack; both items carry negative marginal value (removal improves the stack) and are dropped here.
3. **Contra-Muon → normal ramp extended to step 2500** (`CONTRA_TO_NORMAL_END_STEP: 2000 → 2500`). With Soft-Muon disabled, the late cooldown benefits from Contra-Muon staying active 500 more steps.

Schedule unchanged from PR #294's: `train_steps = 3020`, `schedule_steps = 3050`, PR #287 power-law cooldown (`FINAL_LR_POWER = 1.2`).

## Result

The submitted step count is **2930**. The result directory contains 16 non-cherry-picked seed logfiles for seeds 0 through 15, generated via `torchrun --standalone --nproc_per_node=8 train_gpt_simple.py --seed N` on a single 8×H200 node (Slurm `cluster` partition, `--exclusive`, distinct `--seed N` per run).

At step 2930:

```text
n = 16
mean val loss = 3.27844375
std            = 0.00117834
(3.28 - mu) * sqrt(n) = 0.00622500
```

[...]

PR #307, opened 2026-05-20
Track 3: 2900 steps (n=9) Muon Tail Reference Interpolation

jn2clark:
# Track 3: 2900 steps (n=9) Muon Tail Reference Interpolation

This Track 3 submission keeps the benchmark data stream, global batch size, and
model architecture unchanged. It starts from the PR #300
Aurora-on-`mlp.proj` optimizer stack and re-targets it to an exact 2900-step
endpoint.

The main new endpoint mechanism is Tail Reference Interpolation. The run
captures a full model-weight reference at step `2375`, then evaluates the final
step-2900 checkpoint with:

```text
theta_eval = theta_2900 + gamma * (theta_2900 - theta_2375)
gamma = -0.075
```

Equivalently:

```text
theta_eval = 0.925 * theta_2900 + 0.075 * theta_2375
```

This is a small deterministic interpolation back toward the late-training
trajectory. It is used only at the fixed final validation point and does not use
validation feedback during a run.

## Result

The submitted fixed checkpoint is step `2900`. The nine logs below are clean
full runs from seeds 0 through 8. No run was stopped early, dropped, or selected
based on validation loss.

Under the Track 3 precision rule:

```text
n=9
mean=3.278576
(3.28 - mean) * sqrt(9) = 0.004273
required = 0.004000
```

Using the README's `sigma=0.0013` normal approximation:

```text
z = 3.287179
p = 0.000505982
```

| Seed | Log | 2900 val | Train time | Step avg |
| -: | - | -: | -: | -: |
| 0 | [refinterp_2900_seed0.txt](refinterp_2900_seed0.txt) | 3.27803 | 4798.129s | 1654.53ms |
| 1 | [refinterp_2900_seed1.txt](refinterp_2900_seed1.txt) | 3.27899 | 4778.006s | 1647.59ms |
| 2 | [refinterp_2900_seed2.txt](refinterp_2900_seed2.txt) | 3.27857 | 4907.369s | 1692.20ms |
| 3 | [refinterp_2900_seed3.txt](refinterp_2900_seed3.txt) | 3.27819 | 4755.410s | 1639.80ms |
| 4 | [refinterp_2900_seed4.txt](refinterp_2900_seed4.txt) | 3.27867 | 4842.459s | 1669.81ms |
| 5 | [refinterp_2900_seed5.txt](refinterp_2900_seed5.txt) | 3.27855 | 4767.208s | 1643.86ms |
| 6 | [refinterp_2900_seed6.txt](refinterp_2900_seed6.txt) | 3.27853 | 4695.982s | 1619.30ms |
| 7 | [refinterp_2900_seed7.txt](refinterp_2900_seed7.txt) | 3.27805 | 4861.523s | 1676.39ms |
| 8 | [refinterp_2900_seed8.txt](refinterp_2900_seed8.txt) | 3.27960 | 4814.127s | 1660.04ms |
| **Mean** |  | **3.278576** |  |  |

Run on a single GH200. Edit: confirmed on 8xH100 as well.

## Configuration

Frozen runner:

```text
records/track_3_optimization/results/20260520_tail_refinterp_2900/run.sh
```

Frozen trainer:

```text
records/track_3_optimization/results/20260520_tail_refinterp_2900/train_gpt_tail_refinterp_2900.py
```

Core settings:

```text
TRAIN_STEPS=2900
SCHEDULE_STEPS=3020
CONTRA_TO_NORMAL_END_STEP=2750

TEMPERED_POLAR_RHO=0.05
TEMPERED_POLAR_MAX_DELTA=0.25
TEMPERED_POLAR_RAMP_START_STEP=2600
TEMPERED_POLAR_RAMP_END_STEP=2800

RADIAL_OUTWARD_SCALE=0.5
RADIAL_OUTWARD_SCALE_LATE=0.4
RADIAL_DECAY_START_STEP=2400
RADIAL_DECAY_END_STEP=2900

TARGET_UW_LATE=0.400
TARGET_UW_RAMP_START_STEP=2400
TARGET_UW_RAMP_END_STEP=2800
TARGET_UW_FINAL=0.400
TARGET_UW_SCOPE=all

TAIL_VEL_START_STEP=2500
TAIL_VEL_BETA=0.90
TAIL_VEL_GAMMA=-6.0
TAIL_VEL_MAX_DELTA_RATIO=0.01

REF_EXTRAP_CAPTURE_STEP=2375
REF_EXTRAP_GAMMA=-0.075
REF_EXTRAP_APPLY_START_STEP=2900
```

## Method

This candidate is built on the PR #300 optimizer stack:

- Aurora row-balanced polar update on `mlp.proj`.
- Contra-Muon and the PR294/PR287 optimizer schedule lineage.
- SOAP-style MLP/V preconditioning and radial/u-w-floor machinery inherited
  from the existing Track 3 line.

The exact-2900 changes are:

- Extend Contra-Muon later, to `CONTRA_TO_NORMAL_END_STEP=2750`.
- Use late-only Tempered Polar, ramping `rho` from 0 to `0.05` during steps
  `2600..2800`.
- Retune the late radial brake from `0.5` to `0.4` during steps `2400..2900`.
- Ramp the update/weight floor to `TARGET_UW=0.400` during steps `2400..2800`.
- Keep an EMA of recent parameter updates from step `2500` and apply a clipped
  final tail-velocity transform with `gamma=-6`.
- Capture a step-2375 reference and evaluate step 2900 with the fixed
  `gamma=-0.075` reference interpolation.

What worked:

- Late-only Tempered Polar was better than perturbing the full trajectory.
- The late radial and `TARGET_UW=0.400` retune improved the exact-2900 endpoint.
- Tail velocity with negative gamma helped versus earlier final-weight probes.
- The step-2375 reference interpolation was very helpful, especially on
  weak seeds.

What did not work:

- Positive Tail Extrapolated EMA/XEWA was monotone worse in the seed-1 sweep.
- More Tempered Polar rho/cap surface tuning did not move the mean enough.
- Training-loss EMA gating was too weak to separate good and bad seeds.

## Differences from PR #300

This comparison uses the local PR #300 extraction:

```text
records/track_3_optimization/repro_scripts/pr300_aurora_proj_base.py
```

PR #300 is kept as the base optimizer stack: Aurora on `mlp.proj`, the extended Contra-Muon/Aurora PR294 line, the PR287-style power cooldown, and the existing SOAP/radial/u-w-floor machinery. The exact-2900 candidate changes the endpoint behavior around that base.

Material method changes:

| Area | PR #300 base | Exact-2900 candidate |
| - | - | - |
| Target endpoint | `TRAIN_STEPS=3020`, validated/reporting around the 2930 bin | `TRAIN_STEPS=2900` exact endpoint |
| LR schedule horizon | `SCHEDULE_STEPS=3050` | `SCHEDULE_STEPS=3020` |
| Contra-Muon blend | `CONTRA_TO_NORMAL_END_STEP=2500` | `CONTRA_TO_NORMAL_END_STEP=2750` |
| Tempered Polar | absent | late ramped residual, `rho=0.05`, `max_delta=0.25`, ramp 2600 to 2800 |
| Radial brake | fixed `RADIAL_OUTWARD_SCALE=0.5` | late retune from `0.5` to `0.4`, decay 2400 to 2900 |
| u/w target floor | fixed `TARGET_UW=0.3825` | ramp to/final `TARGET_UW=0.400`, ramp 2400 to 2800, `TARGET_UW_SCOPE=all` |
| Tail velocity final transform | absent | update-EMA from step 2500, `beta=0.90`, final `gamma=-6.0`, clipped at `0.01` parameter norm |
| Final reference interpolation | absent | capture step 2375 and evaluate step 2900 at `0.925 * theta_2900 + 0.075 * theta_2375` |

Harness/probe changes in the experimental trainer:

- Adds final-weight probe hooks for Tail EMA/XEWA, tail velocity, Muon weight scaling, projection scaling, and reference interpolation sweeps.
- Adds single-process checkpoint save/resume support for late-stage screening.
- Adds dense-eval controls and config logging for the 2900 target runs.

Those harness features are not part of the frozen method unless their corresponding config is enabled. The full-log results in this note use only the fixed candidate settings listed above.

## Reproduction

Run the submitted seeds:

```bash
records/track_3_optimization/results/20260520_tail_refinterp_2900/run.sh
```

Run a shorter check:

```bash
SEEDS="0 1" records/track_3_optimization/results/20260520_tail_refinterp_2900/run.sh
```

Recompute the reported statistics:

```bash
python3 records/track_3_optimization/results/20260520_tail_refinterp_2900/summarize_submission_stats.py --step 2900 \
  records/track_3_optimization/results/20260520_tail_refinterp_2900/refinterp_2900_seed0.txt \
  records/track_3_optimization/results/20260520_tail_refinterp_2900/refinterp_2900_seed1.txt \
  records/track_3_optimization/results/20260520_tail_refinterp_2900/refinterp_2900_seed2.txt \
  records/track_3_optimization/results/20260520_tail_refinterp_2900/refinterp_2900_seed3.txt \
  records/track_3_optimization/results/20260520_tail_refinterp_2900/refinterp_2900_seed4.txt \
  records/track_3_optimization/results/20260520_tail_refinterp_2900/refinterp_2900_seed5.txt \
  records/track_3_optimization/results/20260520_tail_refinterp_2900/refinterp_2900_seed6.txt \
  records/track_3_optimization/results/20260520_tail_refinterp_2900/refinterp_2900_seed7.txt \
  records/track_3_optimization/results/20260520_tail_refinterp_2900/refinterp_2900_seed8.txt
```

## Credits

This submission incorporates features from the following previous submissions
and PRs:

- @kumarkrishna PR274 / Skylight-001: NorMuon-lite row/column variance
  normalization, u/w floor postprocessing, and `lr=0.0375` style Muon setup.
- @nilin PR275 and PR291: Contra-Muon / Soft-Muon lineage.
- @samacqua PR278: SOAP preconditioning machinery / MLP SOAP idea.
- @SPThole PR283: Trustlight attention-SOAP path lineage.
- @yash-oai PR287: power-law learning-rate cooldown.
- PR #300: Aurora on `mlp.proj` plus extended Contra-Muon on the PR294 stack.

### DIFF
--- a/records/track_3_optimization/results/20260520_tail_refinterp_2900/PR_SUMMARY.md
+++ b/records/track_3_optimization/results/20260520_tail_refinterp_2900/PR_SUMMARY.md
@@ -1,0 +1,71 @@
+# PR Summary
+
+## Title
+
+Track 3: Tail Reference Interpolation reaches 3.28 in 2900 steps
+
+## Summary
+
+This PR adds the `20260520_tail_refinterp_2900` Track 3 optimization record.
+The method starts from the PR #300 Aurora-on-`mlp.proj` stack and retargets it
+to an exact 2900-step endpoint. The dataset, validation stream, global batch
+size, model architecture, and training loss are unchanged.
+
+The main new mechanism is Tail Reference Interpolation. The run captures a full
+model-weight reference at step `2375`, then evaluates the fixed step-2900
+checkpoint at:
+
+```text
+theta_eval = theta_2900 - 0.075 * (theta_2900 - theta_2375)
+           = 0.925 * theta_2900 + 0.075 * theta_2375
+```
+
+This is a deterministic final-window interpolation back toward the late
+trajectory. It is combined with late-only Tempered Polar, extended Contra-Muon,
+a late radial/u-w retune, and a clipped final tail-velocity transform.
+
+## Result
+
+At the fixed checkpoint `step=2900`, nine consecutive clean full-run seeds
+reach:
+
+```text
+seed 0: 3.27803
+seed 1: 3.27899
+seed 2: 3.27857
+seed 3: 3.27819
+seed 4: 3.27867
+seed 5: 3.27855
+seed 6: 3.27853
+seed 7: 3.27805
+seed 8: 3.27960
+mean:   3.278576
+```
+
+Track 3 precision:
+
+```text
+(3.28 - 3.278576) * sqrt(9) = 0.004273 >= 0.004
+z = 3.287179
+p = 0.000505982
+```
+
+## Reproduction Files
+
+The result folder includes:
+
+```text
+records/track_3_optimization/results/20260520_tail_refinterp_2900/train_gpt_tail_refinterp_2900.py
+records/track_3_optimization/results/20260520_tail_refinterp_2900/run.sh
+records/track_3_optimization/results/20260520_tail_refinterp_2900/refinterp_2900_seed0.txt
+...
+records/track_3_optimization/results/20260520_tail_refinterp_2900/refinterp_2900_seed8.txt
+```
+
+Run:
+
+```bash
+records/track_3_optimization/results/20260520_tail_refinterp_2900/run.sh
+```
+
+The default runner replays seeds `0 1 2 3 4 5 6 7 8` sequentially.
--- a/records/track_3_optimization/results/20260520_tail_refinterp_2900/README.md
+++ b/records/track_3_optimization/results/20260520_tail_refinterp_2900/README.md
@@ -1,0 +1,186 @@
+# Exact-2900 Tail Reference Interpolation
+
+Author: Jesse Clark
+
+This Track 3 submission keeps the benchmark data stream, global batch size, and
+model architecture unchanged. It starts from the PR #300
+Aurora-on-`mlp.proj` optimizer stack and retargets it to an exact 2900-step
+endpoint.
+
+The main new endpoint mechanism is Tail Reference Interpolation. The run
+captures a full model-weight reference at step `2375`, then evaluates the final
+step-2900 checkpoint with:
+
+```text
+theta_eval = theta_2900 + gamma * (theta_2900 - theta_2375)
+gamma = -0.075
+```
+
+Equivalently:
+
+```text
+theta_eval = 0.925 * theta_2900 + 0.075 * theta_2375
+```
+
+This is a small deterministic interpolation back toward the late-training
+trajectory. It is used only at the fixed final validation point and does not use
+validation feedback during a run.
+
+## Result
+
+The submitted fixed checkpoint is step `2900`. The nine logs below are clean
+full runs from seeds 0 through 8. No run was stopped early, dropped, or selected
+based on validation loss.
+
+Under the Track 3 precision rule:
+
+```text
+n=9
+mean=3.278576
+(3.28 - mean) * sqrt(9) = 0.004273
+required = 0.004000
+```
+
+Using the README's `sigma=0.0013` normal approximation:
+
+```text
+z = 3.287179
+p = 0.000505982
+```
+
+| Seed | Log | 2900 val | Train time | Step avg |
+| -: | - | -: | -: | -: |
+| 0 | [refinterp_2900_seed0.txt](refinterp_2900_seed0.txt) | 3.27803 | 4798.129s | 1654.53ms |
+| 1 | [refinterp_2900_seed1.txt](refinterp_2900_seed1.txt) | 3.27899 | 4778.006s | 1647.59ms |
+| 2 | [refinterp_2900_seed2.txt](refinterp_2900_seed2.txt) | 3.27857 | 4907.369s | 1692.20ms |
+| 3 | [refinterp_2900_seed3.txt](refinterp_2900_seed3.txt) | 3.27819 | 4755.410s | 1639.80ms |
+| 4 | [refinterp_2900_seed4.txt](refinterp_2900_seed4.txt) | 3.27867 | 4842.459s | 1669.81ms |
+| 5 | [refinterp_2900_seed5.txt](refinterp_2900_seed5.txt) | 3.27855 | 4767.208s | 1643.86ms |
+| 6 | [refinterp_2900_seed6.txt](refinterp_2900_seed6.txt) | 3.27853 | 4695.982s | 1619.30ms |
+| 7 | [refinterp_2900_seed7.txt](refinterp_2900_seed7.txt) | 3.27805 | 4861.523s | 1676.39ms |
+| 8 | [refinterp_2900_seed8.txt](refinterp_2900_seed8.txt) | 3.27960 | 4814.127s | 1660.04ms |
+| **Mean** |  | **3.278576** |  |  |
+
+## Configuration
+
+Frozen runner:
+
+```text
+records/track_3_optimization/results/20260520_tail_refinterp_2900/run.sh
+```
+
+Frozen trainer:
+
+```text
+records/track_3_optimization/results/20260520_tail_refinterp_2900/train_gpt_tail_refinterp_2900.py
+```
+
+Core settings:
+
+```text
+TRAIN_STEPS=2900
+SCHEDULE_STEPS=3020
+CONTRA_TO_NORMAL_END_STEP=2750
+
+TEMPERED_POLAR_RHO=0.05
+TEMPERED_POLAR_MAX_DELTA=0.25
+TEMPERED_POLAR_RAMP_START_STEP=2600
+TEMPERED_POLAR_RAMP_END_STEP=2800
+
+RADIAL_OUTWARD_SCALE=0.5
+RADIAL_OUTWARD_SCALE_LATE=0.4
+RADIAL_DECAY_START_STEP=2400
+RADIAL_DECAY_END_STEP=2900
+
+TARGET_UW_LATE=0.400
+TARGET_UW_RAMP_START_STEP=2400
+TARGET_UW_RAMP_END_STEP=2800
+TARGET_UW_FINAL=0.400
+TARGET_UW_SCOPE=all
+
+TAIL_VEL_START_STEP=2500
+TAIL_VEL_BETA=0.90
+TAIL_VEL_GAMMA=-6.0
+TAIL_VEL_MAX_DELTA_RATIO=0.01
+
+REF_EXTRAP_CAPTURE_STEP=2375
+REF_EXTRAP_GAMMA=-0.075
+REF_EXTRAP_APPLY_START_STEP=2900
+```
+
+## Method
+
+This candidate is built on the PR #300 optimizer stack:
+
+- Aurora row-balanced polar update on `mlp.proj`.
+- Contra-Muon and the PR294/PR287 optimizer schedule lineage.
+- SOAP-style MLP/V preconditioning and radial/u-w-floor machinery inherited
+  from the existing Track 3 line.
+
+The exact-2900 changes are:
+
+- Extend Contra-Muon later, to `CONTRA_TO_NORMAL_END_STEP=2750`.
+- Use late-only Tempered Polar, ramping `rho` from 0 to `0.05` during steps
+  `2600..2800`.
+- Retune the late radial brake from `0.5` to `0.4` during steps `2400..2900`.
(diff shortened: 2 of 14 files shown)


SPThole:
Really interesting PR overall — the late-stage trajectory shaping ideas are genuinely creative, and I like the focus on exact-endpoint optimization rather than just broad schedule tuning.

The only small philosophical question I had was around the reference interpolation part. The interpolation itself feels reasonably aligned with things like EMA/SWA-style smoothing, so I don’t think it’s inherently outside the spirit of optimization research. The part that feels slightly boundary-pushing is more the idea of explicitly mixing temporally separated checkpoints from the trajectory history.

For example:

0.925 * theta_2900 + 0.075 * theta_2375

is no longer exactly the raw 2900-step optimizer state, but a deterministic transformation of two historical states. Totally fair that this can be viewed as part of the optimizer pipeline, especially since there’s no extra compute, no extra data, and no per-run validation-based checkpoint selection happening.

I think the broader benchmark-design question it raises is interesting though: where should the line eventually sit between “optimizer design” and “trajectory postprocessing”? Especially if future submissions start extending this direction into larger checkpoint soups, trajectory filtering, learned interpolation coefficients, etc.

So not really arguing this is against the written rules — more that it feels like an interesting boundary case that the benchmark may eventually want to formalize more explicitly as these methods evolve.

Very cool work regardless.

jn2clark:
Thanks very much for the feedback @SPThole! I agree on the early/mid improvements but falling behind on the tail. This seems to affect many ideas and implementations, and it has also been very hard to combine methods to remedy this.

I think you raise some fair boundary questions. I'll elaborate a bit more on the original motivation, where this came from, and why I considered it to be part of this track. Hopefully this is useful to other people's ideas as well.

## TL;DR

The TL;DR for the motivation was:

- start from the basic update rule

$$
\theta_{t+1} = \theta_t + \Delta_{t+1}
$$

- think of the update $\Delta_{t+1}$ as the central optimization object, and be somewhat flexible about how that delta is generated;
- use history/memory to lift the optimizer state beyond raw parameter space, in a way that is related to but not limited to standard momentum;
- explore whether the previous trajectory can contribute to extrapolation, trajectory modeling, learned-optimizer-style behavior, or "learn-to-step" type ideas.

Not everything ended up being used as originally intended. The final version here is a much simpler instantiation of that broader idea.

## Background

The framing is that the reference interpolation is a deterministic trajectory transform, or final optimizer readout, rather than checkpoint selection.

Let $\theta_t \in \mathbb{R}^n$ be the parameter vector after optimization step $t$, and define the realized optimizer update as

$$
\Delta_{t+1} = \theta_{t+1} - \theta_t.
$$

For plain SGD,

$$
\Delta_{t+1} = -\eta g_t,
$$

where

$$
g_t = \nabla_\theta L_t(\theta_t).
$$

So under SGD, the weight delta is literally a scaled negative gradient.

For stateful optimizers such as momentum, Adam, Muon-like methods, SOAP/Shampoo-style preconditioners, etc., the update is more generally something like

$$
(h_{t+1}, \Delta_{t+1}) = \Phi(\theta_t, g_t, h_t),
$$

$$
\theta_{t+1} = \theta_t + \Delta_{t+1},
$$

where $h_t$ is optimizer state/history.

In that view, the optimizer is already a dynamical system over an augmented state $(\theta_t, h_t)$, not a memoryless map from the current gradient to the next weights.

The interpolation used here can be written as

$$
\theta_{\mathrm{eval}} = \theta_{2900} + \gamma(\theta_{2900} - \theta_{2375}),
$$

with

$$
\gamma = -0.075.
$$

Since

$$
\theta_{2900} - \theta_{2375} = \sum_{k=2376}^{2900} \Delta_k,
$$

this is equivalently

$$
\theta_{\mathrm{eval}} = \theta_{2900} + \gamma \sum_{k=2376}^{2900} \Delta_k.
$$

Expanding around initialization gives

$$
\theta_{\mathrm{eval}} = \theta_0 + \sum_{k=1}^{2375} \Delta_k + (1+\gamma) \sum_{k=2376}^{2900} \Delta_k.
$$

With $\gamma = -0.075$, this becomes

$$
\theta_{\mathrm{eval}} = \theta_0 + \sum_{k=1}^{2375} \Delta_k + 0.925 \sum_{k=2376}^{2900} \Delta_k.
$$

Although it appears simple, it was not intended to be just "mixing two checkpoints" in an arbitrary way. It can also be viewed as a very simple fixed linear filter on the late-stage optimizer path: the final readout slightly downweights the accumulated displacement from step 2375 to step 2900.

Equivalently, the term

$$
\sum_{k=2376}^{2900} \Delta_k
$$

is the net accumulated displacement over the recent optimizer trajectory. This does not preserve the full path, but it does add a compact memory coordinate beyond the endpoint $\theta_{2900}$. The final readout therefore depends not only on the final weights, but also on a fixed summary of recent optimizer motion.

That is the connection I see to optimizer history and memory.

- Momentum uses a linear recurrence over past gradients.
- Adam maintains moving averages of first and second moments.
- EMA/SWA-style methods maintain a running state over past weights.
- Learned optimizers learn update maps of the form

$$
\Delta_{t+1} = f_\phi(g_t, h_t, \theta_t, \ldots),
$$

with a learned or hand-designed state update

$$
h_{t+1} = U_\phi(h_t, g_t, \theta_t, \ldots).
$$

This PR is obviously much simpler than a full learned optimizer, but I think it sits in the same broad family of ideas: use history/memory of the trajectory to shape the effective update or final readout. In this case the instantiation is deliberately very low capacity: a single stored trajectory anchor and a fixed scalar linear readout.

Another way to phrase this is as a small state-space lift.

Standard training usually returns only the final parameter vector

$$
\theta_T.
$$

Here the optimizer state is augmented with a trajectory memory coordinate. For example, with $t_0 = 2375$ and $T = 2900$,

$$
z_T =
\begin{bmatrix}
\theta_T \\
\theta_{t_0}
\end{bmatrix}.
$$

The submitted weights are then a fixed linear readout from this lifted state:

$$
\theta_{\mathrm{out}} = R(z_T) = \theta_T + \gamma(\theta_T - \theta_{t_0}).
$$

Equivalently,

$$
\theta_{\mathrm{out}} = \theta_T + \gamma \sum_{k=t_0+1}^{T} \Delta_k.
$$

The method is not using validation to choose among checkpoints. It is augmenting the optimizer with a small amount of memory and applying a deterministic linear readout from the lifted trajectory state.

This also has a loose connection to delay-coordinate ideas from dynamical systems. If the current observed state is not sufficient to describe the dynamics, one can augment it with delayed observations. For an optimizer trajectory, a delay-style state over updates might look like:

$$
\xi_t =
\begin{bmatrix}
\Delta_t \\
\Delta_{t-1} \\
\vdots \\
\Delta_{t-m+1}
\end{bmatrix}.
$$

A simple trajectory model could then predict or modify the next update via a linear readout such as:

$$
\widehat{\Delta}_{t+1} = A \xi_t.
$$

That would be a more explicit delay-coordinate-style learned trajectory model. The version in this PR is much more compressed: instead of storing many delayed updates, it stores one trajectory anchor and uses the accumulated displacement:

$$
d_T = \theta_T - \theta_{t_0} = \sum_{k=t_0+1}^{T}\Delta_k.
$$

The lifted state can also be written as:

$$
\tilde{z}_T =
\begin{bmatrix}
\theta_T \\
d_T
\end{bmatrix},
$$

with readout:

$$
\theta_{\mathrm{out}} = \theta_T + \gamma d_T.
$$

This is not a full delay embedding in the formal Takens-theorem sense, but it is motivated by the same broad idea: when the endpoint alone is an incomplete description of the system, augment the state with trajectory history so that a simple readout can use information about recent motion. This is similar in spirit to how momentum, EMA, Adam moments, or learned optimizers enlarge the optimizer state beyond the raw parameters.

## Final thoughts

I would not claim that every weight delta is literally a gradient of some scalar objective. However, I do think weight deltas are valid optimization objects. They are the realized update vectors that define the actual path through parameter space. Under SGD they are exactly scaled negative gradients; under stateful optimizers they are processed, filtered, preconditioned, weight-decayed, or history-conditioned update signals derived from gradients and optimizer state.

Importantly, $t_0 = 2375$ and $\gamma = -0.075$ are fixed as part of the method, not chosen per run from validation behavior. There is no validation-based checkpoint selection, no sweep over candidate anchors at evaluation time, no extra data, and no extra forward/backward training compute.

Hopefully that helps with some of the background and motivation. Let me know if you have any other questions or ideas.

PR #314, opened 2026-05-25
More portability

xor2k:
Unfortunately, I don't have 8x H100s, but two DGX Spark :sweat_smile: Making a PR here about improving the overall portability.

### DIFF
--- a/train_gpt.py
+++ b/train_gpt.py
@@ -32,7 +32,6 @@ import torch.distributed as dist
 import torch.nn.functional as F
 
 # torch._inductor.config.coordinate_descent_tuning = True # we have banned this flag for new records because it causes compilation to take 30min
-from kernels import get_kernel
 from torch import Tensor, nn
 
 from triton_kernels import XXT, XTX, ba_plus_cAA, FusedLinearReLUSquareFunction, FusedSoftcappedCrossEntropy, transpose_add, transpose_copy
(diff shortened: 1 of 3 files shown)


andrewbriand:
I have a couple other portability things I've been wanting to add but haven't made a PR for yet, if you want to include them:

Detect number of GPUs in `run.sh`:
```
#!/bin/bash
NUM_GPUS=$(nvidia-smi --list-gpus | wc -l)
torchrun --standalone --nproc_per_node=$NUM_GPUS train_gpt.py
```

Only run fused RELU kernel on compute capability 90 (e.g. for RTX 4090):
```
major, minor = torch.cuda.get_device_capability(0)
compute_capability = f"{major}{minor}"

class FusedLinearReLUSquareFunction(torch.autograd.Function):
    @staticmethod
    def forward(ctx, x, W1, W2):
        if compute_capability == "90":
            pre, post = linear_relu_square(x.view((-1, x.shape[-1])), W1)
        else:
            pre = x.view((-1, x.shape[-1])) @ W1.T
            post = torch.nn.functional.relu(pre).square()
        x3 = post @ W2
        ctx.save_for_backward(x, W1, W2, pre, post)
        return x3.view(x.shape)

    @staticmethod
    def backward(ctx, grad_output):
        x, W1, W2, pre, post = ctx.saved_tensors
        dW2 = post.T @ grad_output
        if compute_capability == "90":
            dpre = linear_relu_square(grad_output.view((-1, grad_output.shape[-1])), W2, aux=pre)
        else:
            tmp = grad_output.view((-1, grad_output.shape[-1])) @ W2.T
            dpre = 2 * tmp * torch.nn.functional.relu(pre)
        dW1 = dpre.T @ x
        dx = dpre @ W1
        return dx.view(x.shape), dW1, dW2
```

It would also be nice to automatically set `grad_accum_steps` higher for smaller GPUs, but I'm not sure of the best way to do that.

xor2k:
Wow, that's pretty cool stuff, also improves compatibility on my machine and makes a previous workaround of mine (which I did not include in this PR) unnecessary. I've updated this PR.

train_gpt.py
```diff
@@ -1059,7 +1059,36 @@ class AttnArgs:
     xsa_alpha: torch.Tensor | None
     train_max_seq_len: torch.Tensor
 
-flash_attn_interface = get_kernel('varunneal/flash-attention-3').flash_attn_interface
+try:
+    flash_attn_interface = get_kernel('varunneal/flash-attention-3').flash_attn_interface
```
andrewbriand: Does this work on DGX spark for you?
```
flash_attn_interface = get_kernel('kernels-community/flash-attn3', version=1).flash_attn_interface
```
It works on 4090 for me (see https://github.com/KellerJordan/modded-nanogpt/pull/299#issuecomment-4453038117).

xor2k: They only offer x86 builds, can't work on a DGX Spark, which is aarch64. But the question is whether those custom kernels are actually faster/slower or just as fast as the built-in kernels in PyTorch.

xor2k:
Accidentally closed, reopening

PR #317, opened 2026-05-28
New WR (-2.17s): Remove Sparse Attention Gates, Improve XSA Lowering (-17.33s on 1xH100)

aryavohra:
@popov-aleks and I have thought a bit about how [Exclusive Self Attention (XSA)](https://github.com/KellerJordan/modded-nanogpt/pull/264) and [Sparse Attention Gates (SAG)](https://github.com/KellerJordan/modded-nanogpt/pull/117) compose, and concluded it makes sense to disable SAG given XSA.

Both SAG and XSA provide attention-sinking behavior (see [tweet](https://x.com/zhaisf/status/2033692664147939493) from XSA author); we believe that the composition of the two is unnecessary. We have struggled to get 8xH100 allocations, so we ran on a 1xH100 node. We will rerun on 8 GPUs as soon as possible.

Our reproduction of the current record code took 1390 steps to achieve 3.2799 loss in 643.201s. We have been unable to make 1385 steps work for the current record, perhaps the step savings don’t add perfectly across the past two records? Or it could be an issue on our end. In any case, we use this (1390 steps, XSA (@35fbb66), 3.2799 loss) as our baseline.

At 1395 steps, we get an average validation loss of 3.2799 over 10 runs in 625.870s, denoting a runtime reduction 17.33s. If we assume linear improvement by using 8xH100s, this corresponds to ~2.17s. At 1400 steps over those same runs, we get a loss of 3.2786, 629.276s, p-value = 0.01 (~1.74s in 8xH100). We are submitting 1xH100 evidence now and will rerun on 8xH100s ASAP. Thus, disabling SAG yields little meaningful change in the number of steps to hit 3.28 val loss, but yields a runtime improvement through eliminating FLOPS for the redundant SAG logic.

We also tested applying XSA to paired layers, but found this achieves worse results - we haven’t fully understood why, but this implied to us that perhaps paired layers don’t need attention sinks either. Indeed, disabling SAG on these layers too yields a stepwise improvement (and of course a runtime improvement too).

Finally, we applied an algebraic rewrite that causes Inductor to lower XSA better.
```diff
-            vn = F.normalize(v, dim=-1, eps=1e-4)
-            proj = (y * vn).sum(-1, keepdim=True)
+            dot = (y * v).sum(-1, keepdim=True)
+            denom = v.square().sum(-1, keepdim=True).clamp_min(1e-8)
             alpha = torch.tanh(attn_args.xsa_alpha).type_as(y).view(1, 1, self.num_heads, 1)
-            y = y - alpha * proj * vn
-            y = y * torch.sigmoid(F.linear(x[..., :12], attn_gate_w)).view(B, T, self.num_heads, 1)
+            y = y - alpha * (dot / denom) * v
```

We attribute the performance improvement as follows: The algebraic rewrite of XSA does not change the ML nor e.g. FP precision, so isostep runtime improvements ceteris paribus can be attributed to this. It is ~6s (~0.75s on 8xH100), and the other ~11s are from disabling sparse attention gate (~1.4s on 8xH100).


popov-aleks:
We analyzed the function of XSA and confirmed that it induces attention-sink like behavior as the model trains. In particular, the model learns to distribute attention heavily on self-aligned tokens:

Self Alignment Plot

We also performed some causal analysis. Using @_djdumpling XSA architecture trained to 1400 steps (except disabling sparse attention gates to avoid confounding), we performed several eval time perturbations. The two key results are shown below:

| Eval-Time Intervention | $\ell_{\mathrm{eval}} - \ell_{\mathrm{intervention}}$ |
|---|---:|
| No XSA | $-1.7249$ |
| Force $a_{ii}=0$ (Zero out diagonal mass) | $-0.2151$ |

Unsurprisingly, removing XSA leads to massive loss increase, as XSA causes learning of a qualitatively different model.

Perhaps more interestingly, forcing $a_{ii} = 0$, and proportionately redistributing those attention scores to the context tokens before performing XSA leads to only a minor increase in loss. This might suggest attention sinking occurs through the more self-aligned context tokens, as well as the self-token (we have observed some evidence that this is true as the contribution of self-aligned context to attention pre-XSA projection increases after we implement XSA). This might also point to other pathways of XSA action theorized by Zhai such as novelty filtering/removing context redundancy, though we did not test this causally.

popov-aleks:
**UPDATE:** We have now run these improvements on 8xH100, composed with the most recent record from @sisovicm.

The only other code change is to the schedule: I increased the number of scheduled steps to 1390 from 1380 (while keeping the number of extension steps at 10):

num_scheduled_iterations = 1390</code></pre><p>On our hardware, the results of the baseline and XSA-lowering improvement at 1390 steps were:</p><div><div>
variant | n | mean val loss | mean runtime
-- | -- | -- | --
baseline | 6 | 3.2794 | 80.656s
XSA algebraic rewrite only | 6 | 3.2781 | 80.388s

</div></div>
<p>This gives a runtime improvement of <strong>0.268s</strong> from the XSA algebraic rewrite alone.</p><p>Although the new XSA operation is algebraically equivalent, floating-point rounding means the computations can yield slightly different results, even with paired seeds. We did not use paired seeds here, instead averaging over multiple runs. The apparent improvement in mean loss after 1390 steps is likely due to noise. Nevertheless, the runtime improvement appears significant and stable.</p><p><p>For the full change, including removing Sparse Attention Gates, we get:</p>

| n | mean val loss | std | one-sided p-value | mean runtime |
|---|---:|---:|---:|---:|
| 24 | 3.2793 | 0.00124 | 0.00735 | 79.110s |

Thus, at approximately constant validation loss, this version adds **10 extra steps**, but improves runtime by **1.546s**, or about **1.91%**, on the same hardware.

Rough attribution:

| change | runtime improvement | percent improvement |
|---|---:|---:|
| XSA algebraic rewrite | 0.268s | 0.33% |
| remove Sparse Attention Gates | 1.276s | 1.58% |
| total | 1.546s | 1.91% |

The logs are below:
[baseline.zip](https://github.com/user-attachments/files/29434304/baseline.zip)
[xsa_lowering.zip](https://github.com/user-attachments/files/29434306/xsa_lowering.zip)
[this_pr.zip](https://github.com/user-attachments/files/29434307/this_pr.zip)

popov-aleks:
v2: Very small improvement to XSA gate parameterization. Currently, xsa_alphas is initialized as $11 \times H$, but XSA is only active on unpaired attention layers \(1,3,4,7,8,10\). We can store just the six active rows and map them back to their layer positions, removing five unused parameter rows and their optimizer overhead as follows:

`- self.xsa_alphas = nn.Parameter(torch.zeros(num_layers, self.num_heads))`
`+ self.xsa_alphas = nn.Parameter(torch.zeros(6, self.num_heads))`

`- xsa_alphas = [xsa_alpha_per_layer[j] if j in {1, 3, 4, 7, 8, 10} else None for j in range(self.num_layers)]`
`+ xsa_alphas = [None, xsa_alpha_per_layer[0], None, xsa_alpha_per_layer[1], xsa_alpha_per_layer[2], None, None, xsa_alpha_per_layer[3], xsa_alpha_per_layer[4], None, xsa_alpha_per_layer[5]]`

For $n=10$, composed with the above changes, this results in average runtime of 79.006s. The improvement from this change is then 0.104 seconds (0.13%), meaning the improvement from non-SAG changes is 0.372 seconds (0.46%) and the total improvement is 1.65 seconds (2.05%).

ClassicLarry:
Will merge just the XSA rewrite, since the last PR redid the sparse attention gate and XSA layering. Cool stuff though and nice intuition on what to drop. In future, will be fastest if people base their PRs on the latest PR.

-0.3 = 76s.

trianxy:
> In future, will be fastest if people base their PRs on the latest PR.

Makes sense.

Related, I think it would be totally fair if you ask PR authors to rebase their PR on top of the latest PR (or master, if -by then- the latest PR has been merged), before you review it.

ClassicLarry:
> > In future, will be fastest if people base their PRs on the latest PR.
> 
> Makes sense.
> 
> Related, I think it would be totally fair if you ask PR authors to rebase their PR on top of the latest PR (or master, if -by then- the latest PR has been merged), before you review it.

Yea, I'll update readme with this once I go through the stack this week.

PR #322, opened 2026-06-11
Track 1: Recursive solution (leaderboard 79.7s -> 77.34s)

conglu1997:
Track 1 submission from Recursive. This compares against WR #83 re-run on the same 8x H100 Modal setup.

**Results**

| | #83 official | baseline_pr299 (#83, Modal) | this PR (Modal) |
|---|---:|---:|---:|
| n | - | 10 | 13 |
| steps | - | 1385 | 1398 |
| mean val_loss | - | 3.27850 | 3.27893 |
| val_loss std | - | 0.00144 | 0.00137 |
| p(mean<3.28) | - | 0.00471 | 0.007803 |
| mean train_time | 79.7s | 80.61s | 77.34s |
| train_time std | - | 0.22s | 0.20s |

Official PrimeIntellect validation is still pending access to their 8x H100 nodes; these are same-hardware Modal numbers.

**Main changes**

1. FP8 attention projections for QKV and O in the training forward pass. Eval keeps the bf16 path.
2. Annealed row-RMS-scaled Langevin/SGLD exploration noise in NorMuon, zero by cooldown.
3. C-Optim / cautious Adam on the bigram and value-embedding banks.
4. Leaner fused ReLU^2 MLP Triton kernel, storing `post = relu(pre)^2` and reconstructing `sqrt(post)` in backward.
5. Schedule/architecture retunes: fewer paired-head layers, tied embed throughout, and step-count/window changes.

**Rule check**

- Train/val token streams remain FineWeb train/val.
- Mean val_loss passes p<0.01: `p=0.007803` over 13 runs.
- No extra `torch._inductor.config` or compile flags.
- Faster than prior record when baselined on the same hardware: `80.61s -> 77.34s`.

Logs and stats are under `records/track_1_short/2026-06-11_RecursiveFromBest/`.


kroggen:
The title of this PR makes we think that your solution uses some form of recursive (looped/universal) attention

Anyway, very interesting submission! Who would tell that increasing the number of layers would make the training faster?

I suppose it may be required otherwise you would just keep the 10 layers.
Can you explain which of the changes you listed makes the training only beat the record if the number of layers is increased?

conglu1997:
Haha, yes I can see why the name suggests that. We performed leave-out-one ablations on the components and found most of the changes are load bearing at this calibrated step count.

ClassicLarry:
Will merge at 75.4s (-0.6s).

Spent 5 hours or so iterating on these changes, the ReLU^2 MLP Triton kernel holds up and is 0.6s faster. The attn fp8 was giving me nan, and the others were giving increases in loss. Feel free to open a new PR building on the latest + incl other open PRs and validated if want to get those merged in. I think some of the quality changes are not playing nicely with the updated branch, which now has MUDD gates and MLP FP8.

PR #321, opened 2026-06-09
Track 3: aux-β2 + SOAP-f1, clean -- 2750 steps (n=20)

ypwang61:
# Record: Track 3 Optimization -- aux-β2 + SOAP-f1, clean -- 2750 steps (n=20)

## TL;DR

Not a new optimizer: this is the existing **SOAP-Muon** (PR [#278](https://github.com/KellerJordan/modded-nanogpt/pull/278))
with an **auxiliary-β2 split** (longer 2nd-moment memory for the 1-D params — the single biggest lever),
SOAP **extended to all hidden matrices and refreshed every step**, and schedule/momentum tuning, on a
**cleaned-up** stack (seven redundant geometry modules removed).

On **H100 (the benchmark's hardware), n = 20 seeds (0–19)**, it reaches 3.28 in **2750 steps**: mean
3.278856, significance `(3.28 − mean)·√20 = 0.0051 ≥ 0.004` (2745 fails). An independent **n = 8 A40** run
agrees within hardware noise (first-pass 2755; see Result). All while **shrinking the stack to 910 lines**
(smaller and ~19% faster per step), and 2750 is **−25 steps** below the prior 2775 record, at lower loss
and far simpler code.

Relative to the **2875 base (PR [#311](https://github.com/KellerJordan/modded-nanogpt/pull/311),
@liyang2019 — Circuit-Muon on EMA-Nesterov + Aurora)** it applies the local accuracy levers (SOAP on all
hidden matrices + every-step refresh, an auxiliary-β2 split, and schedule/momentum tuning) and **removes
seven redundant geometry modules — including Circuit-Muon itself** — that ablate neutral once the
all-hidden, every-step SOAP is in.

> Note: PR #309/#311 machinery is bundled into the script self-contained, so it reproduces standalone.

full loss curves vs prior records

target zone (zoom)

## Changes (vs the 2875 base, PR #311)

Every constant that differs from the PR #311 code is listed. The parenthetical is the approximate effect
on the first-passing boundary where measured; the removals are loss-neutral (verified identity / within
GPU noise) and their win is **998 → 910 lines (−88) + ~19% faster/step**. Cumulative first-passing boundary
(A40 n=8): **2875 → 2830 → 2800 → 2775 → 2755**; on **H100 (n=20) the boundary is 2750**.

**Added / changed — local accuracy levers:**

[...]

PR #325, opened 2026-06-12
Track3 Tail-EMA 2730 steps (n=8)

jn2clark:
# Track 3: Tail-EMA 2730 steps (n=8)

## TL;DR - cheap tricks still deliver

Not a new optimizer: this is the PR #321 stack (aux-β2 + SOAP-f1 clean,
2750 steps), **byte-identical through step 2400**, plus one mechanism — a
streaming exponential moving average of the weights, blended into the
parameters once at the claimed step. Training stops at 2730 (fixed in advance
for every run; all schedule constants stay at PR #321's values).

**Result: n = 8 non-cherry-picked seeds (0–7), mean 3.278043 at step 2730,
significance (3.28 − mean)·√8 = 0.00554 ≥ 0.004.**

2730 is −20 steps vs the PR #321 candidate (2750).

## Algorithm

The entire change relative to PR #321:

```
EMA_START = 2400     # EMA initialization step
TAU       = 150      # EMA time constant, in steps
LAMBDA    = 0.6      # blend strength at the claim step
S         = 2730     # claim step; training stops here, fixed for all runs
scope     = every parameter except the token embedding

# During training — after the optimizer update of each step t:
if t == EMA_START:
    ema = copy(weights)
elif t > EMA_START:
    ema = ema + (weights - ema) / TAU      # standard EMA update

# Once, immediately before the validation at step S:
weights = (1 - LAMBDA) * weights + LAMBDA * ema
# report val loss at step S; stop.
```

Intuition: under a decaying LR the late iterates orbit the local minimum
rather than sit at it; the EMA averages that orbit noise away
(Polyak/tail averaging). A full replacement (λ=1) hurts because the EMA lags
the still-descending mean, so the optimal blend is partial — λ trades the
EMA's staleness bias against its variance reduction.

## Why these constants are not tuned magic numbers

**τ = (t_end − S) / p.** The natural averaging window for SGD under a
decaying LR is the schedule's local timescale η/|η′|; for the PowerCool
anneal η ∝ (t_end − t)^p (p = 1.2) this gives (t_end − t)/p. We verified the
prediction by measuring the full (τ × λ) surface on two trajectories that
differ only in schedule horizon:

| trajectory | predicted τ* | measured optimum |
|---|---|---|
| t_end = 2900 (submitted) | 170/1.2 ≈ 142 | 150 |
| t_end = 2950 (verification arm) | 220/1.2 ≈ 183 | 200 |

The same verification arm also shows the mechanism is robust to the
trajectory's terminal temperature: the hotter run sits 0.0018 higher in raw
val loss at step 2730, yet after the EMA blend both land within 6e-5
(3.27890 vs 3.27896 on the same seed).

**λ ≈ 0.55.** The last iterate is unbiased but noisy; the EMA is low-variance
but lags the descending mean. The optimal blend of a noisy-unbiased and a
clean-biased estimator is the shrinkage weight λ* = σ²/(σ² + bias²), and when
τ sits at the bias–variance balance point, bias² ≈ σ², so λ* ≈ 0.5.
Measured: 0.5–0.6, flat, consistent across seeds and both trajectories.

## Ablation: against hand-built tail controls

The EMA readout subsumes more elaborate weight-space tail machinery. On the
same stack we first built a two-pulse trajectory-pullback plus a fixed-anchor
readout (in the spirit of the earlier PR #307 reference-interpolation line of
work); it reaches a 2740-step claim (five full official runs included under
`ablation/combo_2740/`, mean 3.278754). Paired same-seed comparison at
step 2730:

| variant | seed 2 | seed 0 |
|---|---|---|
| last iterate (PR #321 unchanged) | 3.28141 | 3.28103 |
| pulses + anchor readout | 3.27983 | 3.27931 |
| **tail-EMA τ=150, λ=0.6 (this PR)** | **3.27876** | **3.27837** |

One mechanism, two constants — both derived above — replaces all of it and
gains a further ~0.0009.

## Reproduction

```
records/track_3_optimization/results/20260611_tailema_2730_submission/train_gpt_tailema_2730.py
records/track_3_optimization/results/20260611_tailema_2730_submission/run.sh
```

The recipe is baked into the script defaults; `run.sh` replays the official
seeds sequentially. Each log embeds the exact source. Per-seed results:

| seed | val @2730 |
|---:|---:|
| 0 | 3.27894 |
| 1 | 3.27813 |
| 2 | 3.27810 |
| 3 | 3.27678 |
| 4 | 3.27900 |
| 5 | 3.27701 |
| 6 | 3.27756 |
| 7 | 3.27882 |

```text
n = 8
mean = 3.278043
(3.28 - mean) * sqrt(8) = 0.005537
required = 0.004
```

Rule-5 note: the stopping step, τ, λ, and the EMA start step are fixed in
advance and identical for every run; no per-run validation-based decisions.
Hardware: 1×GH200 96GB (single-GPU runs; the script runs unmodified on 8×H100
via torchrun nproc 8). GH200 val numbers ran ~+0.0004 above PR #321's
published H100 logs in our paired checks, so this margin is conservative on
the benchmark's hardware.

> Note: PR #321's stack is bundled so the record reproduces standalone; this
> submission builds on (and is −20 steps below) that candidate.


jn2clark:
can take another 10 steps off (2720) as well https://github.com/jn2clark/modded-nanogpt/tree/track3-tailema-2720/records/track_3_optimization/results/20260611_tailema_2720_submission

PR #336, opened 2026-07-11
Improve shifted keys

vfedosov77:
# Training speed improvement: Improved keys offset implementation

## Summary

This change improves training time by optimizing keys offset.

The main idea is: 
if the next token has a weak V, its KV pair is likely underused.
For hub tokens and other overloaded attention channels, shifting K lets the next token reuse the previous token's key, forcing the model to learn a useful V for the next token under that same key.

The adaptive key offset allows reducing training from 1390 to 1370 scheduled iterations while still reaching the 3.28 validation-loss target. The reported wall-clock improvement includes this 20-step reduction.

Hardware: 1xH100

Baseline:
- Mean training time: 834.903 s

With this change:
- Mean training time: 826.562 s

Improvement:
- 8.341 s
- 1.01 %

Runs:

baseline = edf47a05a12062d661c4cfd4eef848c5ab5bed32 (https://github.com/KellerJordan/modded-nanogpt)
modified = b2ba9839c62eb717cd0b6ad87686ea87d4a9ef2c

## Validation loss

losses = 3.2797, 3.2777, 3.2793

Mean loss: 3.2789

The training still reaches the target validation loss of 3.28.

## Reproduction

Nothing is changed - run.sh

### DIFF
--- a/train_gpt.py
+++ b/train_gpt.py
@@ -1091,8 +1091,11 @@ class CausalSelfAttention(nn.Module):
             q, k = yarn.rotary(q), yarn.rotary(k)
 
             if key_offset:
-                # shift keys forward for the stationary head dims. Enables 1-layer induction.
-                k[:, 1:, :, self.head_dim // 2:] = k[:, :-1, :, self.head_dim // 2:]
+                # If the next token has a weak V, its KV pair is likely not used much.
+                # For hub tokens and other overloaded channels, shifting K brings the next channel into the computation.
+                norm_sq = (v * v).sum(dim=-1)
+                shift = (norm_sq[:, :-1] >= norm_sq[:, 1:] * 4.0)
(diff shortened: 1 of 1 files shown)


ClassicLarry:
This is a pretty creative idea. But I am not seeing an improvement when I run this on the current main. Getting 200ms slower and slightly higher less, at equal step count.

vfedosov77:
> This is a pretty creative idea. But I am not seeing an improvement when I run this on the current main. Getting 200ms slower and slightly higher less, at equal step count.

Thank you for the testing! I think some last changes affect the behaviour: previosly it slowed down the training for first ~1100 steps and only then it became much faster and overtake the baseline: this optimization gives hub-tokens a wider channel but it takes time for gradients to discover it.  But maybe now the training is so short that it finishes before the effect takes place:). I will check on the new baseline and try to fix.

PR #337, opened 2026-07-13
Track 1: Prefix token prediction (-0.78s, -15 steps)

jvarho:
Adds another auxiliary target to CE loss, like MTP. However, instead of using a future token, it uses the "prefix token" which is defined here^ as the token corresponding to the longest byte prefix of the current token. Seems to stack with MTP, replacing it was worse, but I did not extensively test different weight combinations.

I have been unable to get my hands on 8xH100, so tests are only on 1xH100, but step time is close enough that I don't think I broke the kernel too badly.

|Run|Steps|N|Time|Loss|p
|---|---|---|---|---|---
|Baseline|1390|5|601190 +/- 639|3.27894 +/- 0.00176|0.14782
|Baseline-|1375|10|594744 +/- 887|3.28091 +/- 0.00238|0.85960
|PTP|1375|10|595214 +/- 642|3.27782 +/- 0.00126|0.00029

This is obviously based on the current MTP implementation and various subtoken/semantic loss ideas from literature, but I'm not aware of this exact form. Let me know if it already has a name.

---

^ Based on tests with a smaller model, data-dependent targets could be even better. E.g. taking the frequency of a token into account when choosing which prefix to use. I'm not sure if that would be within the rules here.


jvarho:
Got access to an 8xH100 and the gains hold:

|Run|Steps|N|Time|Loss|p
|---|---|---|---|---|---
|Baseline|1390|10|79273 +/- 122|3.27874 +/- 0.00236|0.07217
|PTP runs|1375|10|78489 +/- 109|3.27848 +/- 0.00142|0.00539

So this is 0.78s faster on the 8xH100.

varunneal:
Woah this is really cool! Not 100% sure but maybe we have to count the time of the creation of the prefix table in training. Because theoretically you can compute lots of arbitrary language statistics

jvarho:
I figured since it's a static representation of the vocabulary it would fall under init/compile, but I don't mind either way. May need a slightly more optimized implementation in that case.

jvarho:
I smoke tested the merged config with -15 steps and looks like that's still good with the shorter schedule. 3 runs with losses 3.2764, 3.2769, 3.2775. No timing or statistical test, since there are still other PR:s open.

Moving the prefix table generation inside the timed region is a bit annoying since most of the time is taken by tiktoken loading (100+ms). If that can be done outside the loop, then a slightly optimized table build takes 38ms on the instance I had:

```
# Download and load tokenizer
tiktoken.get_encoding("gpt2")

def build_prefix_table(vocab_size: int) -> Tensor:
    byte_to_id = tiktoken.get_encoding("gpt2")._mergeable_ranks
    table = [-1] * vocab_size
    stack: list[bytes] = []  # ancestors of the current token, shortest first
    stack_ids: list[int] = []
    for b, tid in sorted(byte_to_id.items()):
        while stack and not b.startswith(stack[-1]):
            stack.pop()
            stack_ids.pop()
        if stack_ids:
            table[tid] = stack_ids[-1]
        stack.append(b)
        stack_ids.append(tid)
    return torch.tensor(table, dtype=torch.int64)
```

If the tokenizer load isn't allowed ahead of time either then a faster alternative is to read the json manually without loading the full tokenizer (timed at 55ms on my instance):

```
def build_prefix_table(vocab_size: int) -> Tensor:
    GPT2_ENCODER_URL = "https://openaipublic.blob.core.windows.net/gpt-2/encodings/main/encoder.json"
    GPT2_ENCODER_HASH = "196139668be63f3b5d6574427317ae82f612a97c5d1cdaf36ed2256dbf636783"
    encoder_json = tiktoken.load.read_file_cached(GPT2_ENCODER_URL, GPT2_ENCODER_HASH)

    token_to_id = json.loads(encoder_json)
    token_to_id.pop("<|endoftext|>", None)  # special token, no byte prefix: must stay -1
    table = [-1] * vocab_size
    stack: list[str] = []  # ancestors of the current token, shortest first
    stack_ids: list[int] = []
    for token, tid in sorted(token_to_id.items()):
        while stack and not token.startswith(stack[-1]):
            stack.pop()
            stack_ids.pop()
        if stack_ids:
            table[tid] = stack_ids[-1]
        stack.append(token)
        stack_ids.append(tid)
    return torch.tensor(table, dtype=torch.int64)
```

ClassicLarry:
This looks great. Love the idea. Merging at 74.6s, -0.8s and -15 steps. 

I've moved the prefix table into the timing window. Mostly because I'm afraid if we don't, we will end up with a PR that does 20 minutes of tokenizer analysis. Tiktoken loading outside the window is fine.

PR #46, opened 2024-12-07
Replace the fixed 12 by the n_layer in the config

LinglongQian:
Replace the fixed [12](https://github.com/KellerJordan/modded-nanogpt/blob/973030408364f8738b4ad9e8f912d8cbbf56e4d4/train_gpt2.py#L246) and [12](https://github.com/KellerJordan/modded-nanogpt/blob/973030408364f8738b4ad9e8f912d8cbbf56e4d4/train_gpt2.py#L268) by the `n_layer` in the config

PR #339, opened 2026-07-14
Track 3: Bi-Maxwell dual-timescale momentum -- 2635 steps (n=8)

orange4664:
# Track 3: Bi-Maxwell dual-timescale momentum, 2635 steps (n=8)

A 55-step improvement over the current record (#328, 2690 steps), from **one 23-line
change to the momentum's memory kernel**. The improvement is also **pairwise
statistically significant against #46** under the README's pairwise criterion.

## Result

- 8 consecutive seeds (`0-7`) on A800, dense validation every 5 steps over the target zone
- Step **2635** mean validation loss: `3.27852`
- Formal statistic: `(3.28 - mean) * sqrt(8) = 0.00419 >= 0.004` (step 2630 fails with `0.00309`)
- Pairwise vs the #46 logs at the same step: mean gap `0.00357`, giving
  `0.00357 / sqrt(1/8 + 1/8) = 0.00713 >= 0.004`
- **Independent same-recipe fleet on H100 (n=8)**: first formal pass at 2645 (`0.00466`)
- **Pooling all 16 runs** (nothing held back): passes **2635** with margin `0.00487`

Per-seed first crossings (A800): `[2620, 2620, 2640, 2600, 2630, 2585, 2630, 2610]`.

## Change

The #328 stack is retained unchanged (SOAP-Muon, Tail-EMA readout, RowFloor, CWD, radius
pin, EMA-Nesterov, PowerCool LR, mu schedule). One component is added: from step 1000
onward, the Muon first moment's single-EMA memory is replaced by a convex combination of
two fixed-rate EMA buffers (a bi-Maxwell memory kernel):

```python
# per hidden 2-D Muon param, step >= 1000:
M_fast.lerp_(g, 1 - 0.85)                     # fast unit, relaxation ~6 steps
M_slow.lerp_(g, 1 - 0.98)                     # slow unit, relaxation ~49 steps
M_eff = 0.4385 * M_fast + 0.5615 * M_slow     # kernel mean age ~30 steps
update = grad.lerp(M_eff, mu_t)               # scheduled Nesterov mix unchanged
```

At the enable step the buffers lazy-initialize to the current momentum, so that step's
update is bit-identical to the baseline's. Per-step wall-clock is unchanged (two extra
`lerp_`s per hidden matrix, within rerun noise); memory adds two momentum-sized buffers
per hidden matrix.

All three constants were set by measured single-peaked dose-response curves (kernel shape
validated at mean-age parity with two parameterizations; mean-age scan over
{15, 19, 25, 30, 36, 42} peaks at 30; enable-step scan over {0, 500, 700, 1000, 1150}
peaks at 1000). Two controls: with the Tail-EMA readout **disabled**, the same n=8 passes
the formal statistic at raw step 2690 — the raw model matches the previous record's
readout-assisted number, so the gain is in the training dynamics, not the readout; an
annealing control shows the benefit accrues from the mid-training trajectory. Details in
the record README.

The dataset, batch size, architecture, and one-forward/backward-pass-per-step rule are
unchanged.

## Reproduce

```bash
torchrun --nproc_per_node=1 \
    records/track_3_optimization/results/20260713_bimaxwell_2635/train_gpt_bimaxwell_st1000.py --seed 0
```

`summary.tsv` contains the formal aggregate; `A800_seed{0..7}.txt` and `H100_seed{0..7}.txt`
are the full logs with embedded source (all sixteen embed the byte-identical script).

## Curves

Mean validation loss across the 8 A800 seeds vs the #46 record logs (full descent, then the target zone):

mean val loss, full descent

target zone


Nils-p-o:
This is not the place that i expected to see AdEMAMix. I wonder if there is more performance to be squeezed out of the shape of the momentum buffers?

Hongyu-yu:
> This is not the place that i expected to see AdEMAMix. I wonder if there is more performance to be squeezed out of the shape of the momentum buffers?

Our Bi-Maxwell arXiv preprint is online now at:
https://arxiv.org/abs/2608.22994
https://www.alphaxiv.org/abs/2608.22994?utm_source=extension&utm_medium=abstract-page
Hope it interests you.

PR #315, opened 2026-05-27
New track_1 record: MUDD-Gated & Lightweight DC: -60 Steps, -3.51s

Lisennlp:
## V1 Summary

  This PR builds on `2026-05-07_XSAGatedLayers` and combines two changes:

  - Add a lightweight layer-10 DC correction: post-only, no-DD, with FA3 still used for the base
  attention.
  - Add MUDD-generated gates for selected XSA gates, attention gates, x0/bigram injections, and the
  layer-3 skip path.
  - Remove several now-redundant XSA / `attn_gate_w` / fixed injection paths, allowing the schedule to
  drop from `1375 + 10 = 1385` to `1315 + 20 = 1335`.

  ## Results

  Compared with the valid 1385-step XSA baseline:

```text
                   Runs  Steps  Time mean  Time sd  Time delta  Loss mean  Loss sd  Loss p
xsa-baseline-s1385   10   1385    81.2187   0.0681      0.0000     3.2784  0.0013  2.17e-03
xsa-baseline-1335     9   1335    78.5576   0.0401     -2.6611     3.2848  0.0017  1.00e+00
this_pr-s1335        11   1335    78.9669   0.0425     -2.2518     3.2771  0.0012  8.15e-06
```

  - Loss: `3.27841 -> 3.27714`
  - Time: `81.2187s -> 78.9669s` (`-2.25s`)
  - Target test: `p = 8.15e-06`

  At matched 1335 steps, the XSA baseline gets `3.28477`, while this PR gets `3.27714`, for a gap of `0.00763`. DC-only accounts for about `0.0045` loss, leaving roughly `0.0031` from the MUDD-gate / cleanup portion.

  ## Notes

  The DC idea is inspired by DCMHA（[Improving Transformers with Dynamically Composable Multi-Head Attention](https://arxiv.org/abs/2405.08553)）, but this PR uses a lightweight post-only, no-DD variant to control overhead. The current implementation keeps DC separate from FA3, so the custom Triton correction is still relatively expensive; a future FA3-fused implementation should have more optimization room.

PR #346, opened 2026-07-24
New record: 1384 steps / 1.312 min — Ember optimizer on embedding tables + earlier context window extension

katop1234:
# New record: 1384 steps / 1.312 min — Ember optimizer on embedding tables + earlier context window extension

## What changed (one file, +113/−35 vs the current record)

1. **Ember on `value_embeds` and `embed`** ([github.com/katop1234/ember](https://github.com/katop1234/ember)):
   Adam's dense per-coordinate second moment on the two embedding tables is replaced by a
   factored reconstruction `v = outer(r, c) / s` (per-row EMA × globally-pooled per-column
   EMA), with β₁=0 (no momentum buffer). Optimizer state on 46% of the model's parameters
   drops from ~1.86 GB to ~1 MB; distributed cost is one ~768-float `all_reduce` per table
   per optimizer step (~3 KB total) on a dedicated side communicator; stats use
   contiguous reductions only (no atomics), so
   updates are deterministic at fixed world size. `value_embeds` lr ramps 0.64×→1.0× over
   updates 500–690; `embed` runs at a flat 0.5×, cold-started at the untie. `lm_head`
   keeps dense Adam. Measured peak VRAM drops 2.9 GB/GPU (paired runs),
   though at this scale that's a side benefit rather than a speed lever.
2. **Attention windows step to (8,20) at step 1300** instead of at the 1380 extension
   (stage split in `TRAINING_STAGES`; YaRN fires at 1300; the extension-stage transition
   and final-val ws=20 extension become no-ops).
3. **`num_extension_iterations` 10 → 4**: total steps 1390 → 1384. The Muon momentum
   cooldown stays anchored to the 1390-step clock (the run stops partway down the same
   ramp it was tuned on).

## Result (n=20, logs attached, unmodified `run.sh`, byte-identical source in every log)

- Final val loss: **mean 3.27863, sd 0.00157**, 15/20 individually ≤ 3.28 —
  `ttest_1samp(finals, 3.28, 'less')`: **t = −3.89, p = 0.00049** (bar: p < 0.01).
- Timing (measured, not derived): **mean 78.73 s** (min 78.58, max 78.84) across both
  machines. On 8×H100 from PrimeIntellect the mean is **78.68 s vs 78.80 s / 79.10 s for stock on the same box**
  (baseline logs attached in `baseline/`), i.e. faster than the same-hardware stock
  baseline as well as the current record row's 1.320 min. (Comparison base: the 79.2 s /
  1390-step stock code in this repo, i.e. the README row for record 84.)

## Engineering notes

- **State is small enough to skip sharding.** Row stats slice with the shard; column/
  global stats are replicated. Checkpoints no longer depend on world size, and the
  untie's Adam state move (all-gather + transpose + reshard, ~300 MB) is deleted
  outright.
- **NCCL runs collectives in issue order per communicator.** The 3 KB stats all-reduce
  queued behind the 386 MB param all-gathers, costing 2.3%/step in an early version.
  Fix: a dedicated `dist.new_group()` — same bytes, different queue.
- **Deterministic stats.** Contiguous reductions only, no atomics: optimizer state is
  bitwise identical at fixed world size (verified 2-rank == 1-rank). Since every rank
  must hold identical column stats after the all-reduce, a ~3 KB hash doubles as a
  cross-rank desync check.
- **One all-reduce, exact.** The pooled stats are sums of g², which split exactly
  across ranks — a single 3 KB collective gives every rank the exact global statistics.

## Transparency notes

- The 20 logs come from two cloud 8×H100 nodes (10 + 10, both at ~56.9 ms/step reference
  pace, run on 07/18 and 07/20). The embedded source is byte-identical across all 20
  logs and to the file in this PR. The two batches' means differ by 0.23 milli
  (Welch p = 0.75); each 10-run batch alone lands p ≈ 0.011–0.015, so we attach the full
  set rather than a lucky half.
- One run shows a transient val excursion at step 500 (4.45 vs ~4.15 cohort) that fully
  recovers by step 750 and finishes normally; noted for completeness.
- Everything (17-arm sweeps, paired confirms at 1× and 8×, negative results included) is
  documented in the linked repo's findings if anyone wants the full trail.

### DIFF
--- a/README.md
+++ b/README.md
@@ -194,6 +194,7 @@ Note: The 3.28 target was selected to match [Andrej Karpathy's GPT-2 (small) rep
 82 | 1.353 minutes | [Learnable XSA](https://x.com/classiclarryd/status/2058975556520329302) | 04/29/26 | [log](records/track_1_short/2026-04-29_XSAGatedLayers/this_pr_v1-s1410/06563169-6435-48ba-a1ad-f3e61bfcc573.txt),[PR](https://github.com/KellerJordan/modded-nanogpt/pull/264) | @_djdumpling
 83 | 1.328 minutes | [Sign Trick on Bigram Embed](https://x.com/classiclarryd/status/2063061926092099868) | 05/20/26 | [log](records/track_1_short/2026-05-20_BigramsSignTrick/pr299/0cf91274-eda8-49cd-9a97-9369f730f271.txt),[PR](https://github.com/KellerJordan/modded-nanogpt/pull/299) | @TrianX
 84 | 1.320 minutes | FP8 on MLP up-projection forward pass | 05/21/26 | [log](records/track_1_short/2026-05-19_FP8MLPUpProj/this_record/008bb79d-d5bc-4205-bd4e-5e4ae82e658c.txt),[PR](https://github.com/KellerJordan/modded-nanogpt/pull/306) | @sisovicm
+85 | 1.312 minutes | Ember optimizer | 07/20/26 | [logs](records/track_1_short/2026-07-20_EmberWindows) | @katop1234
 ## Rules
 
 New records must:
--- a/records/track_1_short/2026-07-20_EmberWindows/baseline/0931b170-e16b-45cc-9222-99e2e2a845fe.txt
+++ b/records/track_1_short/2026-07-20_EmberWindows/baseline/0931b170-e16b-45cc-9222-99e2e2a845fe.txt
@@ -1,0 +1,4693 @@
+import os
+import sys
+
+# Read the current file and the kernels file code ASAP, for logging
+with open(sys.argv[0], 'r') as f:
+    code = f.read()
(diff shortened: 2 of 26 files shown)


katop1234:
Congrats on #315/#317 — will rebase onto the new record and post updated numbers once the pending window-related PRs settle.

ClassicLarry:
I am going to pass on this PR for now. With only 0.28s gain for a new optimizer, higher loss (requires more than 10 runs to hit p value <0.01), and based on an older branch that would need to be revalidated. The PR drops 6 steps to get the speedup, and its not super clear to me I can't get a similar-ish speedup from dropping 6 steps on the baseline and taking a slightly higher loss. Will come back and revisit at a future time once the open PRs are merged to see if there's a cleaner integration point with bigger wins.

PR #348, opened 2026-07-29
Track 1: Topological Layer Dropout (-0.2, +8 steps)

jon123boss:
## Summary

This PR adapts whole-block layer dropout from [*Don't Drop Dropout: Optimizing Layer Sparsity for Efficient LLM Training and Inference*](https://openreview.net/pdf?id=AEHn9ya4Re) to the current short-track training code. I came across this work while attending ICML 2026. I also tested [AdaRoPE](https://arxiv.org/abs/2607.19363), [Bounded Hyperbolic Tangent (BHyT)](https://arxiv.org/abs/2601.09719), and [NITP](https://arxiv.org/abs/2605.24956). Of those directions, only layer dropout produced a positive result in my testing.

The proposed **topological layer dropout** variant:

- skips at most one complete transformer block per packed-sequence forward
- samples only from the architecture-safe alternating blocks `1,3,5,7,9`
- applies the paper's inverse-density residual scaling, decreasing dropout schedule, and fully active recovery period
- synchronizes sparse events across ranks while rank-offsetting the selected block so every rank executes equal depth
- when layer 3 or 7 is skipped, caches its unchanged input so later MUDD and skip connections still receive the correct residual stream
- precompiles the baseline and all five sparse graphs

Topological layer dropout is enabled by default with `LAYER_DROPOUT_PMAX=0.20`, a 400-step plateau, a 100-step linear decay, and 8 recovery iterations. Set both `LAYER_DROPOUT_PMAX=0` and `LAYER_DROPOUT_RECOVERY_ITERATIONS=0` to reproduce the baseline.

## Baseline, weighted, and topological runs

All reported runs used 8×H100 GPUs on RunPod.

The frozen evidence compares three run families:

- **Baseline** most recent commit [`bc1b58e`](https://github.com/KellerJordan/modded-nanogpt/commit/bc1b58e).
- **Weighted** uses whole-block layer dropout but allocates dropout events unevenly across candidate layers using per-layer weights. It is an experimental control for the allocation strategy.
- **Topological** is the proposed simpler variant. It allocates dropout events equally over the architecture-safe blocks `1,3,5,7,9`, while ensuring every rank executes the same number of blocks.

“Recovery 8” means eight fully active iterations are added after the baseline 1,300-step schedule; “recovery 6” means six are added.

| Configuration | n | Mean loss | Loss SD | Mean time (s) | Time SD (s) | One-sided p vs. 3.28 | Decision |
| --- | ---: | ---: | ---: | ---: | ---: | ---: | --- |
| baseline | 5 | 3.27796 | 0.00203 | 77.0982 | 0.0486 | 0.0438 | control |
| weighted recovery 8 | 5 | 3.27754 | 0.00103 | 76.8566 | 0.0523 | 0.00295 | passes |
| **topological recovery 8** | **5** | **3.27678** | **0.00098** | **76.8804** | **0.0913** | **0.000908** | **proposed** |
| weighted recovery 6 | 5 | 3.28028 | 0.00244 | 76.7208 | 0.0346 | 0.595 | rejected |

8xH100 baseline, weighted, and topological layer-dropout runs

Topological recovery eight saves `0.2178 s` (`0.2825%`) relative to the baseline mean. Across its five frozen runs, the one-sample statistic against the `3.28` loss ceiling is `t(4) = (3.27678 - 3.28) / (0.0009783 / sqrt(5)) = -7.3601`, giving a one-sided `p=0.000908`. This satisfies the required `p<0.01` threshold.

Topological and weighted recovery eight differ by less than the five-run noise. The proposed form therefore uses equal allocation over the topologically safe layers and one scalar scale instead of carrying per-layer weighting machinery.

The compact record in this PR contains the CSV, generated plot, plotting script, and summary.


ClassicLarry:
Its pretty neat you got this to improve loss. I think the amount of code and conceptual complexity here is not worth the 0.2s gain at this time. Will leave it in the open PR list, so if the repo ever stalls out this can get reconsidered, or perhaps it inspires some follow ons.

Issue #88, opened 2025-03-08
Inference performance


mosamdabhi:
Is anyone able to reproduce reasonable inference (next token prediction) performance using the modded nano GPT? Any guide on how the hellaswag performances achieved and in what time?

chnaaam:
I want to know, too

PR #349, opened 2026-08-06
New record: 1.079 min (64.72s) — new ANVIL optimizer + full-stack fp8 and schedule overhaul (−9.3s / −12.6% same-hardware)

devenpzak:
Created by: Deven Pietrzak ([Social Physics Lab](https://socialphysics.inc), MIT Math)

Previous record: 1.23 min (#89).

This run finishes in **64.72s** — the mean of a 12-run, fully logged pool of the exact shipped source on a fresh Xeon 8480+ machine (range 64.56–64.87; logs in the records folder). The original certification pool measured 64.95s on its fastest machine (five runs, 64.87-65.05), 65.36s pooled over all 16 runs.

The certification behind it: sixteen unseeded runs across four machines, every run counted, full tables, including per-machine groupings, in `records/track_1_short/2026-08-08_ANVIL/`. Val CE across the pool averages **3.27604** (σ 0.0009), clearing the 3.28 gate at one-sided p < 1e-10. Per-run logs have since been regenerated as promised: two later log-bearing pools (28 runs on two fresh machines) are in the records entry — the 12-run pool of the exact shipped source averages val **3.2774** (σ 0.0014, one-sided p = 0.00003 vs the gate) at **64.72s** wall. Independent retiming on main is the standard merge gate and reproduces both wall and val on neutral hardware.

For a direct comparison, I re-ran the previous record (#89) in the same sessions: **74.02s** mean over four logged runs on the machine that hosted the shipped-source pool, and 74.38s / 75.05s on two machines of the original pool. That makes this an improvement of **9.30s / 12.6%** on identical hardware, with logs for both sides included.

This submission introduces **ANVIL** (Averaged, Normalized Velocity with Isotropic Lanes), a new optimizer I created.

ANVIL drives each weight matrix with a whitened velocity. Gradients feed twin momentum rails, fast and slow, blended into a Nesterov lookahead. The velocity then passes through a six-map spectral whitening cascade whose coefficients were derived from scratch: its composite envelope is [0.9971, 1.0097] against roughly [0.86, 1.14] for the classical 5-map schedule, about 20 times narrower. Per-lane energy equalization balances the result at fixed Frobenius norm, with sign-aligned decay on the update. A tail-blend ship step folds a tail average of the final weights into the shipped model.

It builds on the line of orthogonalized-momentum optimizers; full lineage citations are in the trainer docstrings.

The gain comes from several places: full-fp8 MLP forward and backward passes with the surrounding kernel work (the largest share), narrow query/key attention (8 of 10 layers compute Q/K at head width 64 with a rotating rotary basis; two full-width carrier layers), the batch/window/schedule design including 40 grown mid-schedule steps, period-4 accumulation cadences on the embedding-table Adam channels (betas squared and weight decay doubled at engagement: the per-update equivalents), a virtual sequence cap (train-time attention masking only; token stream, targets, and validation byte-identical to stock), rewrites of the embedding path (a selected-load value-embedding backward and a sparse-gradient bigram sink), timing and pipeline engineering, and the ANVIL optimizer.

Per review feedback, `train_gpt.py` is modified in place at the repo root. `dc_triton_kernels.py` and `run.sh` are untouched, and the kernel modules new to this submission are additive files at the root; `triton_kernels.py` is no longer imported by this track and is left untouched. The records folder holds the record documentation and statistics, matching the recent record PRs' layout. The four commits separate concerns: the previously submitted version restructured in place with the timing correction applied, the new record, a scrub of inert diagnostics and dead scaffolding (verified no-op at the shipped configuration), and review follow-ups — an fp8 quantizer-state parity fix at the warmup reset (under 1 ms of effect), the per-run logs, and a further dead-code strip. The head commit contains, byte-for-byte, the exact trainer that produced the logged 12-run pool, so a retime can be run directly on it. Also per the earlier timing comment, the prefix-table build, data-shard loading, batch fetching, the rotary-table restore, and the terminal weight-averaging ship steps all run inside the timed section, and the record was retimed on that basis.

PR #344's amax accumulation pattern is compatible with this trainer and measured worth a further ~0.6s at these shapes; deferring to that PR rather than duplicating it here.

Reproduce: `bash run.sh` (env: `KX_STEPS=1178` default; the trainer's baked schedule growth adds 40 (1218 scheduled) plus 23 extension steps = 1241 trained; `DATA_PATH` set to the standard FineWeb10B bins; deps: torch 2.10.0+cu128, triton, `kernels`, tiktoken, numpy). Records run unseeded; per-run numbers in `records/track_1_short/2026-08-08_ANVIL/`.


jvarho:
This moves quite a few things out of the timed region, including the prefix table build that was just moved into it.

devenpzak:
> This moves quite a few things out of the timed region, including the prefix table build that was just moved into it.

Just double checked and our submission runs four things outside the timed region that #89 pays inside it: 1. the prefix-table build (~0.3s depending on implementation). The other three were all noted in the timing note. first-batch fetch + shard prescan (~0.2s), the terminal weight-average blends (~0.02s), and a rotary-table restore (~0.01s). 

The prefix-table placement traces to porting the mechanism from PR #337 as originally submitted, before the build was moved on-clock during that record’s review. I started the build back when that was the current record. Apologies for that. 

I’m moving all four inside the timed section and retiming everything right now. Matching #89’s placement exactly, including filling the prefix table after the clock starts. Will re-certify with a fresh multi-run pool plus fresh same-day #89 anchors on the same machines, then update this PR with revised numbers. Expected impact is roughly +0.5s on our mean, bringing the total to an 8.2s improvement; I’ll report whatever it measures.

jvarho:
> I’m moving all four inside the timed section and retiming everything right now. Matching #89’s placement exactly, including filling the prefix table after the clock starts. Will re-certify with a fresh multi-run pool plus fresh same-day #89 anchors on the same machines, then update this PR with revised numbers. Expected impact is roughly +0.5s on our mean, bringing the total to an 8.2s improvement; I’ll report whatever it measures.

Cool, I really don't know which of those should be timed vs not, the table just seemed like a mistake.

devenpzak:
> > I’m moving all four inside the timed section and retiming everything right now. Matching #89’s placement exactly, including filling the prefix table after the clock starts. Will re-certify with a fresh multi-run pool plus fresh same-day #89 anchors on the same machines, then update this PR with revised numbers. Expected impact is roughly +0.5s on our mean, bringing the total to an 8.2s improvement; I’ll report whatever it measures.
> 
> 
> 
> Cool, I really don't know which of those should be timed vs not, the table just seemed like a mistake.

Yup, makes sense. Table was overlooked due to porting from an old PR before the most recent record (with the table timing change) was merged. 

I'm truly not entirely sure about the other three, but it's probably safer to just include them and err on the side of underestimating the win.

trianxy:
Wow, this is quite a big PR to swallow :) Any chance @devenpzak you can implement this as a minimal delta to the current files of the repo? This would make it easier to review and learn from it.

devenpzak:
> This moves quite a few things out of the timed region, including the prefix table build that was just moved into it.

Fixed and retimed. Prefix-table build, shard loading, first-batch fetch, and terminal weight-averaging ships now all run inside the timed section (table allocated untimed, filled after the clock starts, as with #88). The cost was ~.5s as predicted. I also added a new mechanism, narrow QK-attention. 8 layers compute QK at head-width 64, plus adding steps to recover the loss. The updated number comes out at 64.95s. Details in the updated description

devenpzak:
> Wow, this is quite a big PR to swallow :) Any chance @devenpzak you can implement this as a minimal delta to the current files of the repo? This would make it easier to review and learn from it.

Done, and now in place at the repo root rather than in the records folder.

kroggen:
@devenpzak you posted `train_gpt.py` inside the `‎records/080626_ANVIL/` folder so we cannot see the diff clearly here on GitHub UI

If you pay attention to other submissions, they usually send the `train_gpt.py` file modified in place

Just look at other PRs here

devenpzak:
> @devenpzak you posted `train_gpt.py` inside the `‎records/080626_ANVIL/` folder so we cannot see the diff clearly here on GitHub UI
> 
> If you pay attention to other submissions, they usually send the `train_gpt.py` file modified in place
> 
> Just look at other PRs here

fixed - `train_gpt.py` is now modified in place at the repo root, so the diff renders normally. `dc_triton_kernels.py` and `run.sh` are unchanged, and the new kernel modules are additive files at the root.

varunneal:
Woah! I think this might be quite impressive though I haven't checked it in detail to see if its legitimate. I ran the file through AI and had it check for some common hacks. I found a few that might be suspect 
- FP8 quantizer states should be reset after warmup (I'm not confident which PR introduced this regression)
- some various things are moved out of timing windows, I guess noted above, but like the YaRN Table, BoS scanning, as well as the newly introduced tail-average blend.
 I don't know how big of a difference these make but would be curious on ablations for some weird stuff like deferring python GC until after 100% of training is done, turning off train logs
 
 I think it would be cool to submit the new optimizer (ANVIL) in the Track 3 records to see how it performs against NorMuon, SOAP, etc.

devenpzak:
> Woah! I think this might be quite impressive though I haven't checked it in detail to see if its legitimate. I ran the file through AI and had it check for some common hacks. I found a few that might be suspect
> 
> * FP8 quantizer states should be reset after warmup (I'm not confident which PR introduced this regression)
> * some various things are moved out of timing windows, I guess noted above, but like the YaRN Table, BoS scanning, as well as the newly introduced tail-average blend.
>   I don't know how big of a difference these make but would be curious on ablations for some weird stuff like deferring python GC until after 100% of training is done, turning off train logs
> 
> I think it would be cool to submit the new optimizer (ANVIL) in the Track 3 records to see how it performs against NorMuon, SOAP, etc.

Taking these in order:

**1. The quantizer state.** Real but tiny. It's under 1 ms of skipped arithmetic in a 65 second run, and the one value that could affect training quality came through in a bad state and saturated part of the first optimizer step, so it was slightly hurting this submission, not helping it. Fixed in the latest commit, and I re-ran certification from scratch on the fixed code: per-run logs for two fresh pools are now in the records folder, 12 runs of the exact shipped source at **64.72s mean, val 3.2774** (one-sided p = 0.00003 vs the 3.28 gate) with the current record at **74.02s mean** over four same-session runs on the same machine, a 9.30s / 12.6% gap, plus a 16-run pool on a second machine (65.30s / 3.2770, same-session baseline 74.66s).

**2. Tail-average shipping "after training but before the clock stops":** It already ships inside the clock. The weight-ship steps run at `train_gpt.py:3655-3705` and the final timer read comes after them, at line 3709.

**3. The BoS scan:** inside the clock in both this PR and master. The data loader is a generator; its first `.send()` doesn't execute until step 0, which is on the clock. The `Shard` class is byte-identical to master's.

**4. YaRN tables:** the initial build happens once at model construction, which is untimed in both this PR and master. Every rebuild during training is on the clock in both.

Items 2-4 were true of the first revision of this PR and were fixed in the restructure several commits ago. The diff your AI reviewed no longer exists, so it's worth pointing it at the current head.

**5. GC and logging:** this has the category wrong. Garbage collection isn't training work being deferred, it's avoidable interpreter overhead: `gc.freeze()` removes it outright (frozen objects are never rescanned), the timed section needs no collection to run, and the two lines of setup are available to any submission, including the baseline. The log is one buffered handle with thinned prints, about 3.4 KB past the clock, sub-millisecond. The rules themselves treat systems-side speed work as a normal category (rule 2 even carries an explicit waiver for it), this is disclosed in the records README, and it sits in the code the standard retime times anyway. Worth a few tens of milliseconds against a 9.3 second margin; this trainer holds the record with default GC and stock logging unchanged.

**6. ANVIL on track 3:** haven't thought about this yet, but I may look into it in the near future.

1rreverent:
I don't want to be that guy but if you're going to constantly reply using ai, how would we have a good guess of who we're talking to. 

https://www.pangram.com/history/6d382171-e6db-446c-8e98-56a57c5b46a1?ucc=aBdqHL6jDuD
https://www.pangram.com/history/68aa028c-66f6-4760-954a-cef02ab0e89d?ucc=

If it said 80% it would've been a lot more justifiable

devenpzak:
Closing since #360 replaced this.

PR #350, opened 2026-08-06
Track 1: Canonical token masking (-0.8s?, -15 steps)

jvarho:
I happened to notice that MTP loss pushes probability mass toward non-canonical tokens, so I searched the literature and found: https://arxiv.org/abs/2506.07956
This is basically the local conditioning variant of that paper, masking out non-canonical tokens during inference/validation.

Since this is eval only and provably better for loss, the only question is how many steps can be avoided. I ran it on current master and -15 steps, as well as on #344 and -15 extension steps. Both seem to pass and land roughly where expected:

|Run|Steps|N|Time|Loss|p
|---|---|---|---|---|---
|Baseline|1285|5|547358 +/- 719|3.27836 +/- 0.00208|0.0766
|-masked|1285|5|547358 +/- 719|3.27664 +/- 0.00206|0.0110
|Baseline-|1270|5|540551 +/- 697|3.27848 +/- 0.00072|0.0046
|-masked|1270|5|540551 +/- 697|3.27678 +/- 0.00072|0.0003
|PR-344|1315|5|518020 +/- 844|3.27696 +/- 0.00205|0.0149
|-masked|1315|5|518020 +/- 844|3.27528 +/- 0.00204|0.0033
|PR-344-|1300|5|509029 +/- 961|3.27920 +/- 0.00076|0.0397
|-masked|1300|5|509029 +/- 961|3.27750 +/- 0.00078|0.0010

(Note, these are runs that printed both losses at the end, -0.0017 across 20 runs.)

The table is built before training starts so the intermediate val losses can use it. It's just an optimization, this could be done as part of validation, but it would be slower. The build takes ~3.15s and is CPU only work that isn't needed during training, so if it must be timed I can push it to a background process/thread. I tested that and could not measure any effect on training time, but it's extra complexity, so I left it in a separate branch here: https://github.com/jvarho/modded/tree/canonical-masking-timed

Timing is 1xH100 only so far. This does not touch anything during training so the step count should translate directly.
On current master that would mean -1.2% or -0.8s, on top on PR-344 it would be -1.7% or -1.1s (because of the extension steps). I can try to get a 8xH100 timing when I know which version to test.

### DIFF
--- a/records/track_1_short/2026-08-06-CanonicalMasking/baseline-1270/0afc5496-4f76-4466-b880-35beae7210c6.txt
+++ b/records/track_1_short/2026-08-06-CanonicalMasking/baseline-1270/0afc5496-4f76-4466-b880-35beae7210c6.txt
@@ -1,0 +1,6452 @@
+import os
+import sys
+
+# Read the current file and the kernels file code ASAP, for logging
+with open(sys.argv[0], 'r') as f:
+    code = f.read()
+with open(os.path.join(os.path.dirname(sys.argv[0]), 'triton_kernels.py'), 'r') as f:
+    code += f"\n\n{'-'*40}\n# triton_kernels.py\n{'-'*40}\n\n"
+    code += f.read()
+with open(os.path.join(os.path.dirname(sys.argv[0]), 'dc_triton_kernels.py'), 'r') as f:
+    code += f"\n\n{'-'*40}\n# dc_triton_kernels.py\n{'-'*40}\n\n"
+    code += f.read()
+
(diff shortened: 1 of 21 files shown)


ClassicLarry:
Great find! I went ahead and moved it into the timed region w/ your implementation, and pulled off 5 steps from extension window, for 0.4s improvement. 68.0 -> 67.6s. On latest branch I wasn't seeing low enough loss with 10 steps off. Extension window steps are slower so gain is still pretty good.

PR #353, opened 2026-08-14
Tune AdamW baseline to 4950 steps

konstmish:
## Changes

- `train_steps`: 3250 -> 4950
- Replace Muon on the 2D block-matrix parameters with AdamW
- AdamW learning rates:
  - embedding: 0.7
  - projection: 0.004
  - 1D parameters: 0.015
  - 2D block matrices: 0.001
- AdamW weight decay:
  - embedding, projection, and 1D parameters: 0.001 -> 0.002
  - 2D block matrices: 1.0
- 2D block-matrix AdamW betas: `(0.9, 0.9)`
- Add a 100-step linear warmup from `1e-7`

Validation logs are included in:

`records/track_3_optimization/results/20260701_tuned_adamw_steps_4950_wd_1_wd2_0.002_matrix_b2_0.9/`

## 10-run result at 4950 steps

- mean val loss: 3.277604
- median val loss: 3.277095
- std: 0.002516
- min: 3.274590
- max: 3.282550

## README significance criterion

```text
(3.28 - mean) * sqrt(10) = 0.007577
```

This passes the required `>= 0.004`.

### DIFF
--- a/records/track_3_optimization/results/20260701_tuned_adamw_steps_4950_wd_1_wd2_0.002_matrix_b2_0.9/005c492b-1811-4ae6-ad46-0b364e379fc1.txt
+++ b/records/track_3_optimization/results/20260701_tuned_adamw_steps_4950_wd_1_wd2_0.002_matrix_b2_0.9/005c492b-1811-4ae6-ad46-0b364e379fc1.txt
@@ -20,6 +20,16 @@ import torch.nn.functional as F
 import torch.distributed as dist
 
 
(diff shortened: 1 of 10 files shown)


konstmish:
I was surprised by both the tuned hyperparams and the fact that so many iterations are needed. But it appears at least to be locally optimal, combinations of learning rate and weight decay near the found values did not seem better. Warm-up is kinda optional here, it just makes the final val loss a bit better.

PR #358, opened 2026-08-27
Track 1: Norm CSE, FP8 up-proj quantize (-0.729s)

matevz-kovacic:
## Summary

This PR builds on the current Track 1 `master` record (#89, FP8 MLP down
projection) and adds three changes to `train_gpt.py`, +37 / -5 lines:

**-0.729 s** mean, measured as a paired same-lease comparison against the prior
record on two independent 8xH100 leases (6.0x and 8.7x the measured noise floor).
**20 independent runs: mean val loss 3.27886, p = 0.001408.**

The standard run command is unchanged:

```bash
python data/cached_fineweb10B.py 10
torchrun --standalone --nproc_per_node=8 train_gpt.py
```

## What Changed

- **CSE of `norm(cache[7])`.** `cache[7]` is written once and `norm` is pure, so
  layers 8/9/10 each recomputed the same full-size RMS norm. Hoisted into a
  per-forward memo: two redundant norms and their backwards removed.
  Value-identical.
- **Fused FP8 re-quantize of the MLP up-projection.** The up-projection has run in
  FP8 since record 84; this does not change that, it fuses the re-quantize. The
  down-projection got a fused path at record 89; the up-projection did not,
  because it lives in a module method `torch.compile` does not reach. A compiled
  helper fuses the divide and the cast, so the full-size bf16 intermediate is
  never materialised. Same arithmetic.
- **`num_extension_iterations` 15 -> 10.** These run after the cooldown completes,
  at final lr and window, so trimming them does not compress the decay schedule.
  Their measured loss cost is not negligible -- see below.

The first two are systems-only. The third changes a step count and so consumes
loss buffer, which is why rule 2's `p < 0.01` is met here rather than waived.

## Main Result

Timing is reported as a **paired same-lease comparison** — the prior record and
this PR on one machine in one session, with the baseline bracketed before and
after the candidate so mid-lease drift is visible. Absolute seconds are never
compared across machines.

| | prior record (same lease) | this PR | margin | 2*SE noise floor |
|---|---:|---:|---:|---:|
| lease A, n=10 each | 74.9270s | 74.2178s | **-0.7092s** | 0.1190s (**6.0x**) |
| lease B, n=10 each | 74.6815s | 73.9323s | **-0.7492s** | 0.0863s (**8.7x**) |

Baseline drift within lease A was +0.0891s and within lease B +0.0336s, both well
below the margin.

| | this PR |
|---|---:|
| n | 20 |
| Mean val loss | 3.2788600000 |
| Loss std | 0.0014869255 |
| p(mean < 3.28) | **0.0014077373** |
| Mean train time | 73.345350s |
| Time std | 0.142198s |

Logs and per-run statistics are in
`records/track_1_short/2026-08-27_NormCSE_FP8UpQuant_ExtTrim/`
([README](https://github.com/matevz-kovacic/modded-nanogpt/blob/normcse-fp8up-exttrim/records/track_1_short/2026-08-27_NormCSE_FP8UpQuant_ExtTrim/README.md)).

## The Loss Buffer (rule 2's discretionary clause)

Change 3 reduces a step count, so the discretionary clause applies and I should
show it beats a simple step-count decrease at equivalent loss. **I cannot show
that, and I would rather say so than argue it.**

What is measured, on this codebase:

| change | val CE cost | wall-clock |
|---|---:|---:|
| cut 40 **scheduled** steps (1270 -> 1230) | +0.00671 | -2.32s |
| cut 80 **scheduled** steps (1270 -> 1190) | +0.01050 | -4.63s |
| cut 5 **extension** steps (15 -> 10) | +0.00099 (t = 1.82, p = 0.09, n.s.) | -0.31s |

The extension figure is a paired same-lease comparison at n=10 vs n=9 against the
same candidate with the tail left at 15. The scheduled figures come from a
4-point sweep over four hosts at n=4 with a fixed-seed repeat harness.

Scheduled steps cost 1.7e-4 val CE each, so a scheduled cut costing the same
0.00099 would buy about 0.34s against the extension trim's 0.31s. Those are
indistinguishable, and the comparison crosses measurement tiers, so it does not
establish superiority in either direction.

Two things I can state without hedging. Scheduled-step reduction is not a route
to multi-second gains here: 1230 lands 17.9 sigma over the 3.28 gate, so the
record's step count is genuinely tuned. And the extension trim's cost is not
statistically significant while the resulting candidate still clears p < 0.01 on
20 independent runs.

**If you would rather not spend buffer at all, changes 1 and 2 stand alone at
-0.396s on the certification lease** (five independent leases, mean -0.365s),
both value-identical or same-arithmetic, zero buffer consumed, rule 2 waived as
systems-only. I am glad for that to be the record instead -- say the word and I
will drop change 3.

## Validity

**Target p-value passes.** `p = 0.0014077373` for mean val loss < 3.28, below the
Track 1 `p < 0.01` requirement.

**Same data and batching.** Train and validation token streams are unchanged
(rule 1). The data pipeline, batch size and sequence configuration are untouched.

**Same command and compile settings.** No extra `torch._inductor.config` or
`torch.compile` flags (rule 3). The one `@torch.compile` added is a decorator on a
new helper, matching the decorators already used throughout the file.

**Independent runs, not repeats.** Each of the 20 runs is a separate `torchrun`
process with fresh initialisation and no seeding, so the spread is genuine
inter-run variance. Two collected logs were excluded on stated criteria before any
statistic was computed — one preempted at step 1205/1280, one whose final-step
validation never reached the file — by log completeness, not by value.

**Same hardware comparison.** 8x H100 80GB SXM, Nebius eu-north1, PyTorch
2.10.0+cu128, Triton 3.6.0, CUDA 12.8, driver 570.211.01, base commit `75d6867`.
Our nodes run slower than the reference host — our re-run of the current record
reads 74.93s, not 73.80s — so every claim here is a same-hardware delta. Official
PrimeIntellect validation is pending.

**Measured as a set.** The first two changes were measured together; separated,
neither clears the noise floor at n=10 (+0.134s and +0.177s against floors of
0.235s and 0.179s).

## Run Data

This PR losses:

```python
[3.2785, 3.2777, 3.2793, 3.2777, 3.2802, 3.2786, 3.2798, 3.2769, 3.2793, 3.2783, 3.2798, 3.2820, 3.2760, 3.2773, 3.2795, 3.2795, 3.2800, 3.2777, 3.2813, 3.2778]
```

This PR times:

```python
[73.250, 73.253, 73.621, 73.381, 73.648, 73.286, 73.308, 73.294, 73.164, 73.336, 73.306, 73.300, 73.227, 73.423, 73.310, 73.207, 73.263, 73.372, 73.292, 73.666]
```

```python
import scipy.stats
print("p=%.10f" % scipy.stats.ttest_1samp(this_pr_losses, 3.28, alternative="less").pvalue)
```

## Disclosure

Found by **Active Model** ([link](https://github.com/matevz-kovacic/active-model)),
an AI system, working on its own from a single prompt:

> Maximally optimize Modded-NanoGPT Track 1 under the current rules:
> https://github.com/KellerJordan/modded-nanogpt; use README.md for the GPU
> interface, economical GPU usage, reconnaissance information, and a hard GPU
> budget of $300.

It chose the targets, wrote the patches and ran the measurements. One correction
worth stating: the loss claim was re-measured on independent initialisations after
it emerged the first p-value had used same-seed repeats, which moved the margin
from 0.00205 to 0.00114. The reported figure is the corrected one.

Suggested attribution, if accepted:

    @matevz-kovacic and AI System [Active Model](https://github.com/matevz-kovacic/active-model)

## Leaderboard Row

I have not edited `README.md`, since records add the row separately. Filled in so
nothing needs guessing -- `90` assumes this lands next (89 is current head as I
write), so bump it if another record merges first:

```
90 | 1.222 minutes | Norm CSE, fused FP8 up-projection quantize, and trimmed extension tail | 08/27/26 | [log](records/track_1_short/2026-08-27_NormCSE_FP8UpQuant_ExtTrim/this_pr/9683b30c-fe93-5760-8358-e2360fff1ff6.txt),[PR](THIS_PR_URL) | @matevz-kovacic and AI System [Active Model](https://github.com/matevz-kovacic/active-model)
```

`1.222 minutes` is our host's average and is **not** comparable to the existing
column — replace it with whatever PrimeIntellect measures. The linked log is the
run closest to the mean on both val loss and train time, not the best of the 20.


ClassicLarry:
Im not seeing a measurable speedup from the norm change- this is something I had also tested when I added the cached activation. I think pytorch is smart enough to handle it. The MLP change looks good, but unfortunately has been made redundant by one of the PRs that was raised before this. Closing the PR.

PR #342, opened 2026-07-18
Track 1: Add FP8 MLP down projection (-1.04s)

Mister-dev-oss:
## Summary

This PR builds on the current Track 1 `master` record and adds one systems-side
change to `train_gpt.py` and `triton_kernels.py`:

**FP8 MLP down projection.** The MLP down-projection forward path now runs in
FP8 via `torch._scaled_mm`. The existing fused up-projection/ReLU-squared Triton
kernel also emits FP8 post-activation values and partial activation amax values.
The MLP down-projection weights are quantized and transposed to FP8 with delayed
scales. The backward path still uses the original BF16 activations and BF16
weights.

The standard run command is unchanged:

```bash
python data/cached_fineweb10B.py 10
torchrun --standalone --nproc_per_node=8 train_gpt.py
```

## What Changed

- Add FP8 post-activation emission to the fused up-projection/ReLU-squared
  Triton kernel.
- Add delayed activation-scale reduction for the FP8 MLP down-projection input.
- Add FP8 quantize+transpose for the MLP down-projection weights.
- Route the forward down projection through `torch._scaled_mm`.
- Keep the backward path in BF16, using the same saved BF16 tensors as before.

## Main Result

Both baseline and this PR were run on the same 8xH100 setup with the same
environment:

- PyTorch `2.10.0+cu128`
- Triton `3.6.0`
- CUDA `12.8`
- 13 runs per variant

| Run | Mean val loss | Loss std | p(mean < 3.28) | Mean train time | Time std |
| --- | ---: | ---: | ---: | ---: | ---: |
| Current `master` baseline | 3.2782461538 | 0.0020164389 | 0.0042976481 | 80.728615s | 0.065165s |
| This PR | 3.2783692308 | 0.0014156180 | 0.0006690604 | 79.684923s | 0.182767s |

Logs and per-variant statistics are included in:

- `records/track_1_short/2026-07-17_FP8DownProjection/baseline/`
- `records/track_1_short/2026-07-17_FP8DownProjection/this_pr/`

## Delta

Relative to the same-hardware baseline:

- Mean train time: `79.684923s - 80.728615s = -1.043692s`
- Relative train time: `-1.292841%`
- Speedup: `1.01309774x`
- Mean val loss: `3.2783692308 - 3.2782461538 = +0.0001230769`

The loss distributions are statistically indistinguishable in the baseline-vs-PR
comparison:

- Welch p-value for candidate loss vs baseline loss: `0.8587465`

The timing separation is stable across the 13 runs:

[...]

PR #360, opened 2026-08-31
New Record: 0.665 minutes (39.9 seconds): ANVIL2, Sampled-softmax, New Embedding Table, Full-stack fp8 (-34.0s, -46% same hardware)

devenpzak:
Created by: Deven Pietrzak (MIT Math. X: [DevenPzak](https://x.com/DevenPzak). Lab: [Hyperstition (fka Social Physics Lab)](https://x.com/hyperstition_cc))

After a few weeks since my last record without touching the PR, I finally got access to some more H100s. The following record was built over 5 days. **Independently reviewed and reproduced by four separate sources over the past two days**

This PR will replace #349.

| | runs | wall (training section) | final val CE |
|---|---|---|---|
| **this PR** | **18** | **39.914 +/- 0.120 s** | **3.27731 +/- 0.00099** |
| record #89, same machine, in between record runs | 9 | 73.889 +/- 0.137 s | 3.27828 +/- 0.00205 |
| **delta** | | **-33.98 s, -46.0 %, 1.85x** | |

Vs 3.28: p = 9.4e-10**. Cold cache for all. 

Note: Lost one leg's raw logs when the pod I was renting expired. Recorded the result and it's included in stats and footnoted in statistics.md

## Ablation table

Mostly n=2. 

| group | net seconds |
|---|---:|
| **Sampled-softmax training loss** | **+8.40** |
| **Embeddings**, hashed n-gram table | **+7.05** |
| **ANVIL**, optimizer | **+5.08** |
| &nbsp;&nbsp;├─ update rule + anneal/shipping | +4.16 |
| &nbsp;&nbsp;└─ optimizer-update graph capture | +0.92 |
| **Full-stack FP8** | **+4.96** |
| **Mixed-width attention** | **+3.82** |
| **Training-step graph capture** | **+2.99** |
| **Sparse grad comms & data loader** | **+1.20** |
| **MUDDformer residual mix** | **+0.48** |
| **Total** | **+33.98** |

ANVIL is an improvement of 21 millinats at roughly equal wall clock, compared to record 89's shipper optimizer.

**For this table, I estimated the val-to-wall-clock conversion using the exchange rates in ms/millinat that made the changes add up to 33.98s, which ended up as 164 ms/millinat. This was within the empirical measured cost of 120-250 ms/millinat.

## Reproducing it

```bash
bash run.sh
```

Notes for reproducing it in README, but make sure to follow the following especially, as other ways could throw off the reproduction: use the pinned torch build, Driver >= 580, bare metal needs `libcudart.so.13`, make sure to get the kernel properly.

Note:

A lot of people seem to take issue with the bigram table. Wanted to address a couple misconceptions about it:

1. It's only 7 seconds of 34. I've also had people tell me my ablation was wrong, so I went back, removed the embedding table increases, and added steps until the model was back under 3.28. On net, it cost 7 seconds, validating the ablation table. 

2. People think nobody had done it because it was too many parameters / shouldn't be allowed. In fact, it’s the opposite. The table was already 290M parameters (over twice the size of the model), and the reason it hadn’t grown was because it would have made the time worse. 

The old code kept copies of the table on every GPU, wrote and zeroed a table-sized gradient every single step, and wrote the table back to every GPU after every step. This made the cost linear in table size but the gains roughly logarithmic. 290M was the crossing point, which is why it was set to that size. 

I sharded the table across 8 GPUs, pulled only the rows the current batch’s tokens actually hashed to, sent gradients back as a segment-sum rather than table-sized tensor, and ran Adam only on the touched rows. Then the optimizer cost scaled with rows used per step, making it almost invariant to table size (1/4 size embedding table, 16B instead of 65B, only saved 0.17 seconds). This let me push to 224x and add a trigram channel. 

Even if I didn’t expand it, just from the engineering described in the last paragraph, I would have saved over one second from this change alone. The real reason no one had expanded the embedding table was not because of the table itself but rather because it would have been too slow. The table was already significantly larger than the model, so the bigram table was already far too large to ever scale to frontier training.


bai634140-ui:
wow，really?

xTimeCrystal:
65B embedding parameters! I wonder if it is legal to train an [infini-gram](https://arxiv.org/abs/2401.17377) model on CPU first, then achieve the loss req on step 1?

ShizukaKuze:
> 65B embedding parameters! I wonder if it is legal to train an [infini-gram](https://arxiv.org/abs/2401.17377) model on CPU first, then achieve the loss req on step 1?

There's no way this follows the rules. Firstly, the repository maintainer already said, "This is the most tricky; I would say that hardcoding it ok only when there's a coherent theory for why it makes sense, i.e., when we could expect the hardcoded value to generalize to other settings in some sense. Which isn't the case for the lambdas; since I have no idea how to determine them a priori without just doing a run."

I honestly have not looked through the code, but I'm just willing to go on a tangent and say that if there really are 65 billion embedding parameters, it'sunctionallylookup table.

> Disproportionately degrades the readability of the codebase. A 200 line kernel to drop 300ms is considered worthwhile. 500 lines that convolute the optimizer layout for a 50ms gain will likely be rejected.

This is ridiculous. It adds nearly 200,000 lines of code. (I was actually wrong here. Still, 65 billion embedding table kind of defeats the purpose. At that point, how much of the model is actually GPT?

devenpzak:
@xTimeCrystal @ShizukaKuze 

I sincerely appreciate y'all bringing these up, and that's why I included the bigram/trigram count, so this could be discussed. I expanded the already-existing table from 377,280 to 84,602,880 rows. I knew people would ask about it, but I firmly believe that it's legal since it's (a) the same accepted mechanism widened and given a new trigram channel, (b) entirely learned on clock, with substantial collisions preventing memorization, and (c) seems to be clearly allowed by the rules. **It's also only ~20% of the PR's record**. Details below. Sorry for the delay, this took a while to reply to.

**Regarding "nearly 200,000 lines of code"**

> > Disproportionately degrades the readability of the codebase. A 200 line kernel to drop 300ms is considered worthwhile. 500 lines that convolute the optimizer layout for a 50ms gain will likely be rejected.
> 
> This is ridiculous. It adds nearly 200,000 lines of code. (I was actually wrong here. Still, 65 billion embedding table kind of defeats the purpose. At that point, how much of the model is actually GPT?

Of the +194,329 in the diff, **189,260 are the 30 files in the record folder**, which is the attached evidence, solely consisting of 26 run logs, their statistics, and the README. Those are required attachments, not code. If you take a look at past PRs, you'll notice they had similar structure. For example, the most recent two accepted PRs, #342 and #337, had diffs of 125k and 211k lines, respectively. Of course, the actual code diff was similarly smaller, but submitting records inflates the number.

The true code change is **+5,068 / −2,465 across 9 files**, but −1,316 of the deletions is removing `dc_triton_kernels.py`. For a record that's a larger percentage improvement than the last 45 world records combined, I think the 5k is justified. It's likely I could clean it up a bit more with some work, but as of now the code is thankfully not anywhere near the insane diff that 200k would be.

**Regarding the embedding table**

**TLDR**, the table is learned *entirely inside the timed run* and generalizes a previously several-times-accepted mechanism. It's compression, not storage. Even if it was deemed illegal, the published ablation table shows it would take *only 7 seconds* away, leaving us with a 27 second world record improvement.

Now to go through your concerns one at a time: 

> > 65B embedding parameters! I wonder if it is legal to train an [infini-gram](https://arxiv.org/abs/2401.17377) model on CPU first, then achieve the loss req on step 1?

**On pretraining an n-gram model off the clock:** this record does not do that. The table is initialized to zero and learned 100% on the clock, within the timed run:

```python
# train_gpt.py:2183 (this PR)
self.bigram_embed = nn.Embedding(_bgv4_rows, args.bigram_dim,
    _weight=torch.zeros(_bgv4_rows, args.bigram_dim, dtype=torch.bfloat16, device=device))
```

Nothing is precomputed. The untimed warmup ends in the exact same reset as the previous records. And the trainer *asserts* the shard is zero and restores it, for added safety. 

For the specific rules question posed by @xTimeCrystal, I think this has already been ruled on.
@samacqua has a test-time-training run (#205) in the "Notable attempts" section. The description says

> While technically a valid probability model, **we are not allowing untimed backward passes.**

So pretraining an infinigram model falls into the same category of untimed training work. However, my record keeps *everything* related to the table on the clock.

> 
> There's no way this follows the rules. Firstly, the repository maintainer already said, "This is the most tricky; I would say that hardcoding it ok only when there's a coherent theory for why it makes sense, i.e., when we could expect the hardcoded value to generalize to other settings in some sense. Which isn't the case for the lambdas; since I have no idea how to determine them a priori without just doing a run."

**On the hardcoding rule:** It shouldn't apply. Hardcoding here applies to values from past runs or from the val set. The table is solely *learned parameters* trained by Adam on the same training data inside the timed section. There's no hardcoded constant anywhere in the table. And `post_lambdas` and `resid_lambdas` are `nn.Parameter` in record 89 (lines 1334, 1338), which means they're learned, not hardcoded, just like this table.


> I honestly have not looked through the code, but I'm just willing to go on a tangent and say that if there really are 65 billion embedding parameters, it's functionally a lookup table.

**On the model being "functionally a lookup table":** it can't be. The index is a *hash* and the collisions are huge. This is WR 89's exact hash, but I made it wider:

```python
rand_int_1, rand_int_2 = 36313, 27191
mod = args.bigram_vocab_size - 1
out[1:] = torch.bitwise_xor(rand_int_1 * out[1:], rand_int_2 * out[:-1]) % mod
```

Each pair of two token IDs map to one row. There are 2.53 billion possible bigrams, while the bigram table has 42.3 million rows. That means **every row shares ~60 token pairs** (while record 89 shares ~6700). This means we have 60 conflicting gradients, so only common language patterns survive. The model learns real structure, exactly the goal of the speedrun.

There's a difference between storage and compression. If it was an infinigram, every n-gram would have a slot to retrieve exactly, which is memorization. A hash at 60:1 destroys n-gram detail by construction and keeps only shared structure that the model can actually learn. 

Plus, the sign trick has been in many other records. each pair also draws a random ±1 pattern from a separate 8,192-row table on a different hash. This tries to make colliding pairs cancel instead of accumulating. Even the comment on record 89 says *"Use sign-trick to better compress multiple bigrams into a shared bigram embedding row."* which means this method has been both discussed and approved; I just added more rows and a trigram channel.

Also, it's not the output head either. The hashed vector goes into the residual stream, before layer 0, behind a learned gate:

```python
x[..., :args.bigram_dim] = x[..., :args.bigram_dim] + x0_bigram * bigram_gates[0]
```

The transformer still makes every prediction. Plus, the target is held-out validation. Which means only learned structure survives. The model picks up coarse n-gram English statistics from the same token stream, which is the reason the table was introduced in PR 62. 

**On whose mechanism this is:** it was not at all an original change from me. Bigram hash embedding was record 62 (#201, @ClassicLarry), 27 records ago, and it has been in every single record since then. It was moved to GPU in record 68 (#216), given sparse gradient comms in record 71 (#221), and given the sign trick in record 83 (#299). WR 89 has the same module, zero initialization, sign table, Adam channel, sharded-sparse comms, betas, and weight decay as my PR. So this exact mechanism, just with fewer rows, has been intentionally approved four times and accepted in 27 consecutive PRs.

```python
# ecbb586:train_gpt.py:1329-1332
self.bigram_embed = nn.Embedding(args.bigram_vocab_size, args.bigram_dim)
nn.init.zeros_(self.bigram_embed.weight)
bigram_sign_table = torch.randn(args.bigram_sign_table_rows, args.bigram_dim).sign()...

# ecbb586:train_gpt.py:2034
"bigram_embed":   {"optim": "adam",    "comms": "sharded_sparse",
                   "adam_betas": [0.75, 0.95], "lr_mul": 75.,  "wd_mul": 5.0},
```

Effectively, all I did was widen the row count and add a trigram channel. In my opinion, it's also noteworthy that the previous table is already ~290M against a 124M transformer. The record is already majority embedding.

**On "how much of the model is actually GPT?"** I'd argue pretty much all of it, for the reasons above. However, it's the maintainers' call officially (@ClassicLarry). Either way, I can put a floor on it, since the published ablation has the entire table at **+7.05s of the 33.98s**. If I had to remove it *entirely*, it would still be ~26.9s faster than WR 89. The transformer is the same, the table costs 0 FLOPs, and no single GPU even holds it (sharded one eighth per rank).

**Regarding whether this whole thing is legal in the first place**

The rules are pretty clear. Rule 1 names architecture as changeable. The rules are data, loss bar, inductor flags, and faster on the same hardware. It follows them up with 

> Other than that, **anything and everything is fair game!**

Later, the target is defined as a valid probability model. It doesn't restrict the composition of the probability model. 

> The target metric is *cross-entropy loss on the FineWeb val set*. To speak mathematically, the goal of the speedrun is **to obtain a probability model of language** which assigns a probability of at least `math.exp(-3.28 * 10485760)` to the first 10,485,760 tokens of the FineWeb valset. Hence, e.g., we allow evaluation at any sequence length, **so long as we still have a valid probability model of language.**

Because of this, "how much of it is GPT" isn't required. The rules are, same tokens, valid probability model, under 3.28, faster on the same box. This PR meets all four.

Hopefully this answers y'all's concerns!! Please let me know if there's anything else you see. This PR makes some large changes, which is why I wanted to have four separate sources review it for both legality and timing. This exact issue was brought up, but the conclusion was that, for the reasons above, this is an acceptable change. There were several other changes, arguably justified by past records and worth a couple seconds, that I decided seemed too questionable to submit.

trianxy:
@devenpzak, wow, this is a big one :)

Personal opinion: It would be of great help if you can create a new sub-PR (or just carve out a first commit) which implements only one part of this PR (eg the ANVIL2 optimizer), with as few code line changes to current master as possible.

Once we understand and confirm the latter diff, I think we can go faster through the rest (and resolve discussions like the new embedding table).

Btw: you cite #89 a lot, but that seems to be a typo

ShizukaKuze:
> h



> @xTimeCrystal @ShizukaKuze
> 
> I sincerely appreciate y'all bringing these up, and that's why I included the bigram/trigram count, so this could be discussed. I expanded the already-existing table from 377,280 to 84,602,880 rows. I knew people would ask about it, but I firmly believe that it's legal since it's (a) the same accepted mechanism widened and given a new trigram channel, (b) entirely learned on clock, with substantial collisions preventing memorization, and (c) seems to be clearly allowed by the rules. **It's also only ~20% of the PR's record**. Details below. Sorry for the delay, this took a while to reply to.
> 
> **Regarding "nearly 200,000 lines of code"**
> 
> > > Disproportionately degrades the readability of the codebase. A 200 line kernel to drop 300ms is considered worthwhile. 500 lines that convolute the optimizer layout for a 50ms gain will likely be rejected.
> > 
> > 
> > This is ridiculous. It adds nearly 200,000 lines of code. (I was actually wrong here. Still, 65 billion embedding table kind of defeats the purpose. At that point, how much of the model is actually GPT?
> 
> Of the +194,329 in the diff, **189,260 are the 30 files in the record folder**, which is the attached evidence, solely consisting of 26 run logs, their statistics, and the README. Those are required attachments, not code. If you take a look at past PRs, you'll notice they had similar structure. For example, the most recent two accepted PRs, #342 and #337, had diffs of 125k and 211k lines, respectively. Of course, the actual code diff was similarly smaller, but submitting records inflates the number.
> 
> The true code change is **+5,068 / −2,465 across 9 files**, but −1,316 of the deletions is removing `dc_triton_kernels.py`. For a record that's a larger percentage improvement than the last 45 world records combined, I think the 5k is justified. It's likely I could clean it up a bit more with some work, but as of now the code is thankfully not anywhere near the insane diff that 200k would be.
> 
> **Regarding the embedding table**
> 
> **TLDR**, the table is learned _entirely inside the timed run_ and generalizes a previously several-times-accepted mechanism. It's compression, not storage. Even if it was deemed illegal, the published ablation table shows it would take _only 7 seconds_ away, leaving us with a 27 second world record improvement.
> 
> Now to go through your concerns one at a time:
> 
> > > 65B embedding parameters! I wonder if it is legal to train an [infini-gram](https://arxiv.org/abs/2401.17377) model on CPU first, then achieve the loss req on step 1?
> 
> **On pretraining an n-gram model off the clock:** this record does not do that. The table is initialized to zero and learned 100% on the clock, within the timed run:
> 
> ```python
> # train_gpt.py:2183 (this PR)
> self.bigram_embed = nn.Embedding(_bgv4_rows, args.bigram_dim,
>     _weight=torch.zeros(_bgv4_rows, args.bigram_dim, dtype=torch.bfloat16, device=device))
> ```
> 
> Nothing is precomputed. The untimed warmup ends in the exact same reset as the previous records. And the trainer _asserts_ the shard is zero and restores it, for added safety.
> 
> For the specific rules question posed by @xTimeCrystal, I think this has already been ruled on. @samacqua has a test-time-training run (#205) in the "Notable attempts" section. The description says
> 
> > While technically a valid probability model, **we are not allowing untimed backward passes.**
> 
> So pretraining an infinigram model falls into the same category of untimed training work. However, my record keeps _everything_ related to the table on the clock.
> 
> > There's no way this follows the rules. Firstly, the repository maintainer already said, "This is the most tricky; I would say that hardcoding it ok only when there's a coherent theory for why it makes sense, i.e., when we could expect the hardcoded value to generalize to other settings in some sense. Which isn't the case for the lambdas; since I have no idea how to determine them a priori without just doing a run."
> 
> **On the hardcoding rule:** It shouldn't apply. Hardcoding here applies to values from past runs or from the val set. The table is solely _learned parameters_ trained by Adam on the same training data inside the timed section. There's no hardcoded constant anywhere in the table. And `post_lambdas` and `resid_lambdas` are `nn.Parameter` in record 89 (lines 1334, 1338), which means they're learned, not hardcoded, just like this table.
> 
> > I honestly have not looked through the code, but I'm just willing to go on a tangent and say that if there really are 65 billion embedding parameters, it's functionally a lookup table.
> 
> **On the model being "functionally a lookup table":** it can't be. The index is a _hash_ and the collisions are huge. This is WR 89's exact hash, but I made it wider:
> 
> ```python
> rand_int_1, rand_int_2 = 36313, 27191
> mod = args.bigram_vocab_size - 1
> out[1:] = torch.bitwise_xor(rand_int_1 * out[1:], rand_int_2 * out[:-1]) % mod
> ```
> 
> Each pair of two token IDs map to one row. There are 2.53 billion possible bigrams, while the bigram table has 42.3 million rows. That means **every row shares ~60 token pairs** (while record 89 shares ~6700). This means we have 60 conflicting gradients, so only common language patterns survive. The model learns real structure, exactly the goal of the speedrun.
> 
> There's a difference between storage and compression. If it was an infinigram, every n-gram would have a slot to retrieve exactly, which is memorization. A hash at 60:1 destroys n-gram detail by construction and keeps only shared structure that the model can actually learn.
> 
> Plus, the sign trick has been in many other records. each pair also draws a random ±1 pattern from a separate 8,192-row table on a different hash. This tries to make colliding pairs cancel instead of accumulating. Even the comment on record 89 says _"Use sign-trick to better compress multiple bigrams into a shared bigram embedding row."_ which means this method has been both discussed and approved; I just added more rows and a trigram channel.
> 
> Also, it's not the output head either. The hashed vector goes into the residual stream, before layer 0, behind a learned gate:
> 
> ```python
> x[..., :args.bigram_dim] = x[..., :args.bigram_dim] + x0_bigram * bigram_gates[0]
> ```
> 
> The transformer still makes every prediction. Plus, the target is held-out validation. Which means only learned structure survives. The model picks up coarse n-gram English statistics from the same token stream, which is the reason the table was introduced in PR 62.
> 
> **On whose mechanism this is:** it was not at all an original change from me. Bigram hash embedding was record 62 (#201, @ClassicLarry), 27 records ago, and it has been in every single record since then. It was moved to GPU in record 68 (#216), given sparse gradient comms in record 71 (#221), and given the sign trick in record 83 (#299). WR 89 has the same module, zero initialization, sign table, Adam channel, sharded-sparse comms, betas, and weight decay as my PR. So this exact mechanism, just with fewer rows, has been intentionally approved four times and accepted in 27 consecutive PRs.
> 
> ```python
> # ecbb586:train_gpt.py:1329-1332
> self.bigram_embed = nn.Embedding(args.bigram_vocab_size, args.bigram_dim)
> nn.init.zeros_(self.bigram_embed.weight)
> bigram_sign_table = torch.randn(args.bigram_sign_table_rows, args.bigram_dim).sign()...
> 
> # ecbb586:train_gpt.py:2034
> "bigram_embed":   {"optim": "adam",    "comms": "sharded_sparse",
>                    "adam_betas": [0.75, 0.95], "lr_mul": 75.,  "wd_mul": 5.0},
> ```
> 
> Effectively, all I did was widen the row count and add a trigram channel. In my opinion, it's also noteworthy that the previous table is already ~290M against a 124M transformer. The record is already majority embedding.
> 
> **On "how much of the model is actually GPT?"** I'd argue pretty much all of it, for the reasons above. However, it's the maintainers' call officially (@ClassicLarry). Either way, I can put a floor on it, since the published ablation has the entire table at **+7.05s of the 33.98s**. If I had to remove it _entirely_, it would still be ~26.9s faster than WR 89. The transformer is the same, the table costs 0 FLOPs, and no single GPU even holds it (sharded one eighth per rank).
> 
> **Regarding whether this whole thing is legal in the first place**
> 
> The rules are pretty clear. Rule 1 names architecture as changeable. The rules are data, loss bar, inductor flags, and faster on the same hardware. It follows them up with
> 
> > Other than that, **anything and everything is fair game!**
> 
> Later, the target is defined as a valid probability model. It doesn't restrict the composition of the probability model.
> 
> > The target metric is _cross-entropy loss on the FineWeb val set_. To speak mathematically, the goal of the speedrun is **to obtain a probability model of language** which assigns a probability of at least `math.exp(-3.28 * 10485760)` to the first 10,485,760 tokens of the FineWeb valset. Hence, e.g., we allow evaluation at any sequence length, **so long as we still have a valid probability model of language.**
> 
> Because of this, "how much of it is GPT" isn't required. The rules are, same tokens, valid probability model, under 3.28, faster on the same box. This PR meets all four.
> 
> Hopefully this answers y'all's concerns!! Please let me know if there's anything else you see. This PR makes some large changes, which is why I wanted to have four separate sources review it for both legality and timing. This exact issue was brought up, but the conclusion was that, for the reasons above, this is an acceptable change. There were several other changes, arguably justified by past records and worth a couple seconds, that I decided seemed too questionable to submit.

In my opinion, if everything holds to be true, this is a very impressive pull request. However, the embedding table made me unironically spit out my drink. If it's truly 65 billion parameters, and if it get merged, I don't think this repository would have very much real world implication anymore. It's obviously up to the project maintainer, and it's just my opinion. But I do strongly believe that if the 65 billion parameter embedding table was merged into this repository, it would seriously degrade this repository's real-world implications. Besides that, I think it's very commendable what you have accomplished!

devenpzak:
> @devenpzak, wow, this is a big one :)
> 
> Personal opinion: It would be of great help if you can create a new sub-PR (or just carve out a first commit) which implements only one part of this PR (eg the ANVIL2 optimizer), with as few code line changes to current master as possible.
> 
> Once we understand and confirm the latter diff, I think we can go faster through the rest (and resolve discussions like the new embedding table).
> 
> Btw: you cite #89 a lot, but that seems to be a typo

Thanks for the message! I will consider working on a sub-PR in the near future, but at the moment I am working on a couple other projects (and currently don't have H100s). Whenever I get a chance I will do this, or hopefully it will already be reviewed by then.

I fixed the PR description - PR 89 was a typo, was supposed to be record 89. That was a stand-in description because I was trying to push it out as soon as possible; now I fixed it and made it a lot more concise.

devenpzak:
> In my opinion, if everything holds to be true, this is a very impressive pull request. However, the embedding table made me unironically spit out my drink. If it's truly 65 billion parameters, and if it get merged, I don't think this repository would have very much real world implication anymore. It's obviously up to the project maintainer, and it's just my opinion. But I do strongly believe that if the 65 billion parameter embedding table was merged into this repository, it would seriously degrade this repository's real-world implications. Besides that, I think it's very commendable what you have accomplished!

That's a fair point, and I appreciate your compliments! 

However, I do feel like this repository's real-world applications don't exclusively come from the entire configuration being deployable in real LLM training. The challenge is more about speed optimization. Of course, that often leads to innovations like Muon (and hopefully now ANVIL) which come out and into frontier usage. But at the same time, things like the specific batch / schedule tuning, disabling garbage collection, the untimed warmup pass, and even the former ~290M parameter bigram embedding table are optimizations that have played a role in dozens of records but have little applications to anything else. The challenge combines both of these into one question of pure optimization, so even if a massive bigram/trigram embedding table doesn't have as many real-world applications, it doesn't mean that other innovations coming from this challenge, such as Muon or ANVIL, can't. This record itself proves that; it saves 34 seconds, but even if you take away the table, it still beats the record by 27 seconds, primarily through real, useful innovations.

BurnyCoder:
https://x.com/vvvincent_c/status/2096275917316522245 
> the most suspicious change is using a 65B-parameter lookup table for training the 124M param model.

Yeah. No.

KellerJordan:
Personally I would consider all of the changes described here to be 100% legitimate. Both @ClassicLarry and I have the power to accept new records— for me to accept I’d need to first reproduce this result myself, & I haven’t had time to do so yet.

ClassicLarry:
I haven't gone through all the specifics yet, but at a high level I agree it looks great. Hoping to find a block of time soon to look at more closely and refactor the code into something easier to follow, now that we are jumping to 5k+ lines. I'd be curious to understand where the intuition came from on Sampled-softmax training loss and Embeddings, hashed n-gram table. And if any of the results are a joint effort across Social Physics Lab or Autoresearch harness on Codex (in the past groups have set records then after the fact revealed a larger initiative behind it, which is totally fine too). In any case, very impressive!

overrule:
Hi @ClassicLarry, I'm the founder of [Hyperstition](https://x.com/hyperstition_cc/status/2098471087415992455) (formerly Social Physics Lab). The nanoGPT speedrun was a testing ground for ANVIL I, the first iteration of our optimizer. Since then, we have released ANVIL III, which beats muon by 20-28 millinats, regime dependent, [tested up to 1.2B scale](https://hyperstition.cc/cutting-pretraining-costs-by-62-percent). Just today, we released an API for our model, [Feather](https://x.com/hyperstition_cc/status/2099972576005214469), which beats Qwen3-1.7b using 180x fewer total training tokens.  

Would love to chat in private, and share some of our findings on broader pretraining research. We'll be further be releasing our results on frontier RL over the coming weeks. For anyone interested in learning more about our work, or simply chatting, drop in a text at aj@socialphysics.inc.

ClassicLarry:
Ok this one is up next. I'll try to merge it in with the latest main, going piece by piece to validate the impact. Hopefully later this weekend. and I'll message you to chat!

PR #363, opened 2026-09-03
Track 1: FP8 MLP input gradient and kernel tuning by AI System Warpscale (-0.45s)

isubuz:
Three changes.

- `linear_relu_square_kernel`: `BLOCK_SIZE_K` 128 to 64 on the FP8 forward. At
  one byte per operand that fits seven pipeline stages under the 232 448-byte
  sm_90 ceiling, where 128 fits three. The BF16 backward is untouched, since
  seven does not fit there.
- `ce_fwd_bwd_kernel`: minimum blocks per SM 2 to 3, softmax-sum loop unrolled
  25 times instead of 2. Output is bit-identical.
- MLP up-projection input gradient in FP8 via `torch._scaled_mm`, delayed
  per-tensor scaling. The Relu^2 kernel writes the FP8 gradient and its partial
  amax during the pass it already makes, so the scale costs no extra read.
  `dW1` and `dW2` stay BF16 because they need a transposed FP8 copy, which
  costs more than the GEMM saves.

13 runs a side, back to back on the same 8x H100 node.

| | baseline | this PR |
| --- | ---: | ---: |
| mean train time | 73.704 s | 73.250 s |
| mean val loss | 3.27845 | 3.27782 |
| one-sided p vs 3.28 | 0.0019 | 0.0086 |

Train time falls 0.4535 s, 19.9 sigma. The arms do not overlap: fastest baseline
73.590 s against 73.316 s for the slowest run here. Val loss falls 0.00064, so
none of the buffer below 3.28 is spent. The p-value gap is one run at 3.2861,
which the record README covers.

Logs, per-arm statistics, and notes on how this was found and what was tried and
rejected are in `records/track_1_short/2026-09-03_FP8MLPBackwardAndKernelTuning/`.

Found with [Warpscale](https://warpscale.ai) driving Claude coding agents.


isubuz:
supporting deep dive blog post here - https://warpscale.ai/blog/deep-dive-into-modded-nanogpt/

